PR #7, opened 2022-08-11
Juno py3 mark

markshowalter:
Extended notes...

1. Your commenting style separates comments from code more cleanly than mine, which is fine (although I am not always clear on the distinction between "++++++" and "======" and "------" and "******" and "######"). However, your lowest level of separator is this:
            #- - - - - - - - - - - - - -
I find this style of separator _extremely_ visually distracting---in the same way that polka dots are distracting. For me, these "polka dot" separators make the code harder to read rather than easier. I urge you to stop using these!

One alternative would be to replace them with the next higher level of separator:
            #---------------------------
My preferred solution would be to remove these separators completely. When I see comments isolated between solid lines of characters, I tend to interpret them as a header, put there to separate larger logical blocks within the code. However, sometimes a comment is there just to explain the following couple of lines, in which case the extra horizontal separators seem like overkill to me. Most of the occurrences of "#- - - - - -" seem to be around this latter kind of comment.

I am not saying that you should change these in the current pull request, but I do ask you to discontinue use of "#- - - - -" going forward, and maybe remove them when you revisit a file in which they appear.

(By the way, does your editor color-code your source? If it doesn't, you should really consider finding one that does. When I look at a Python file, all comments are in red so they are quite easy to distinguish from executable code. Maybe consider PyCharm or Emacs, both of which are available on Linux. Both of them also expand tabs automatically too.)

2. Your editor is also leaving behind lots and lots of trailing whitespace. That in itself is not a big deal, but git recognizes any change in trailing whitespace as a change to the file, so it flags it. This could make pull requests harder to review. I have a setting on my editor to automatically strip trailing whitespace from my Python files when I save. I wonder if you do as well. In any case, try to avoid trailing whitespace, as is mentioned in the RMS style guide.

3. When I revisited all the Observation classes, I was struck by how difficult it is to keep them all straight in my head. To help out, I checked in a new file:
        oops/obs_/NOTES-ABOUT-OBSERVATION-CLASSES.txt
It's a text file describing the distinctions between the various subclasses. I hope it is helpful.

4. After our extended discussions about the Observation and Cadence classes, I came up with an approach that I hope you will find satisfactory. One of the key problems with replacing the timing inputs to all Observation subclasses with a single input (tuple, dict, or cadence) is that it was an invitation to let the user pass in a nonsensical cadence--one that is functionally incompatible with the Observation subclass to which it was applied.

4.1 I created a new Cadence subclass TdiCadence, which handles the timing of TDI cameras. Check it out. (It is not yet well tested so please debug and fill in some unit tests.) By the way, part of the reason for this addition is that the Sequence subclass you were using has some features designed under the assumption that its time steps are consecutive and not overlapping, which TDI cameras violate. (I had built the Sequence subclass for occultations with non-uniform time intervals.) For the Pushframe constructor, you can provide a tuple or dictionary of inputs to the TdiCadence constructor in place of the object itself; this is functionally equivalent to what you had, although some of the parameter names are different.

4.2 I also created a couple of new constructors for the most common observation cadences:
        DualCadence.for_array2d
        Metronome.for_array1d
Now, in any place where the expected input is a Cadence object, you can provide a tuple or dictionary in its place. However, instead of each Observation subclass having its own "default cadence", it will always be one or the other of these. As you note, the tuple/dictionary option can, in some cases, make it easier for the user to build the object they need based on parameters at their fingertips (exposure times, interline delays, etc.) without quite so much knowledge of the subtleties of the Cadence class.

Note, however, that there are unavoidable inconsistencies among the Observation subclasses about the relationship between the "shape" of an FOV and the "shape" of a cadence. This affects which parameters can appear in a given tuple or dict. This is explained in each docstring and also mentioned in the NOTE file.

4.3 I still disallow the specification of a cadence for the Snapshot and Slit1D constructors, because neither of these has any internal time-dependence. My reasoning remains that there is no benefit to the introduction of this option, it invites nonsensical inputs, and a user of one of these subclasses would plausibly ask, what is a "cadence" and why should I care?

Warning: In support of these revisions, there are a lot of changes to the constructors for the Observation subclasses, so look over my changes carefully. They all pass their unit tests currently, but that doesn't mean I didn't overlook something.

5. This is fun... I added an attribute "snapshot" to the 2-D observation subclasses other than Snapshot itself. In each case, it is a Snapshot object that has the same start time, end time, and FOV as the associated object. This is a step toward the implementation of more accurate "inventory" methods for 2-D observations with time-dependence. For now, the inventory method of each of these subclasses is implemented by calling the associated snapshot's inventory method. It's a starting point, and probably not a bad guess for the Pushbroom, Pushframe, and Rasterscan classes. It'll need more work for some of the others.

6. I deleted the sweep_duv_dt method from the Observation class. We don't (yet) use it and I had not thought it through carefully. For one thing, it was bound to have serious divide-by-zero errors as implemented. At some point I thought it would have a role in some derivative calculations, but I am not so sure.

7. In reviewing the Observation subclasses, I noted that I had not followed the Python standard that the names of class attributes only used internally should begin with an underscore. I fixed that. Going forward, there are probably a lot of other places where it makes sense to insert underscores in front of attribute names, so be on the lookout for that.

8. I did not understand what you were trying to do with the "tfrac" attribute of the Pushframe class. I think it might reflect a misunderstanding of what this input to the "timegrid" method does. We can discuss, but for now it is commented out.

9. Note that there are no Pushframe unit tests.


jnspitale:
1. Ok, it sounds like my tolerance for - - - is slightly higher than yours.  The idea is indeed to separate logical blocks of code, so I'd like to keep doing that, but I'm sure there's another way that will work for both of us.  My preferred editor (nedit) has code recognition features, but I have not made use of them because my commenting style predates that technology.  I would suggest that at the - - - level, we dispense with the separators and just rely on the editor coloring, which I have turned on.

2. I have noticed some trailing whitespace as well.  I'm not sire what's causing that, but nedit does have a setting to show white space, so I turned that on and that should make the trailing whitespace obvious.

3. This is very useful.  I would argue for putting it in observation.py as part of the docstring for the Observation class instead of in a separate file.  It would then be available via __doc__ and help().

4. Ok cool, I'll re-work the PushFrame class with this cadence.

5. Ok, this may be another way implementing the snapshot output that I currently have in the jiram module.

8. That may be a residual, I'll take a look

9. The unit tests are there, but I apparently have them bypassed for some reason

PR #13, opened 2022-09-14
Fix various oops bugs (RF+MS+JS)

rfrenchseti:
- Fix some of the bugs caused by Mark's recent "save state" checkin.
- Fix some bugs found by a Python linter in Visual Studio Code.
- Fix some bugs found by running the Cassini navigation code.
- Reorganize and optimize import statements.


rfrenchseti:
The following bugs are known to still exist in oops/polymath and I don't know how to fix them. Many of these are new, but a few have been there for a long time.

- oops/backplane/unittester.py:53

        if zoom != 1:
            image = zoom_image(image, zoom, order=1)

There is no "zoom_image" function anywhere in the pds-oops repo.

- oops/backplane/body.py:396

The function "__setstate__" is full of references to "body" but there is no "body" defined.

- oops/backplane/event.py:203

The function "__getstate__" references ARR_VEC3_PROPERTIES and DEP_VEC3_PROPERTIES, which are not defined.

- oops/fov/tdifov.py: 254

The function "uv_from_xyt" is not complete.

- oops/frame/__init__.py: 918

The function "__getstate__" references PICKLE_CONFIG, which is not defined.

- oops/frame/__init__.py: 1248

The function "_interpolate_matrix_omega_old" references the "utils" module, which is not defined. If this function is old and no longer needed, it should be removed. Otherwise add an import for utils.

- oops/frame/spinframe.py:82

The function "__setstate__" references "axis", which is not defined.

- oops/frame/twovectorframe.py:114

The function "__setstate__" references "pole", "retrograde", "aries", and "cache_size", which are not defined.

- oops/gravity/oblategravity.py:85

The function "potential" references "a2", which is not defined.

- oops/observation/pushframe.py:200

The function "uvt_range" references "uv_max", which is not defined.

- oops/surface/polarlimb.py:132

The function "vector3_from_coords" references "z" and "clock", which are not defined.

- polymath/matrix3.py:168

The functions "x_rotation", "y_rotation", and "z_rotation" reference "self" but these are static methods.

jnspitale:
- What's the rationale for deleting unittester_backplanes.py from backplane/ instead of from oops/?
- in scalar.py, the spirit of this change looks correct, but it doesn't look like it will work because bool() only takes one argument.  Is the purpose here to "or" the nonnegative test with any existing discriminant mask?

rfrenchseti:
OK, you're right that unittester_backplanes.py (and the associated diff file) should probably be in the backplane directory instead of the module root.

rfrenchseti:
scalar.py is totally the wrong fix. Good catch.

rfrenchseti:
On further examination, I don't know the right way to fix the problem with scalar.py. It's a circular import. Scalar wants to use the Boolean class, but Boolean is a type of Scalar, so it isn't defined yet. I'll let Mark look at this one.

hosts/newhorizons/__init__.py
```diff
@@ -51,9 +51,9 @@ def initialize(asof=None, time=None, meta=None):
             spicedb.open_db()
 
             names = spicedb.furnish_lsk(asof=asof)
-            names = spicedb.furnish_pck([5,9] + lrange(501,505) +
-                                                lrange(514,517) +
-                                                lrange(901,906), asof=asof)
+            names = spicedb.furnish_pck([5,9] + list(range(501,505)) +
```
rfrenchseti, line 54: Why were these changed? `lrange` should be proper here.

oops/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
rfrenchseti, line 1: We should add `.vscode` to `.gitignore` and remove this file since we will all have our own preferences.

hosts/juno/__init__.py
```diff
@@ -1,3 +1,535 @@
 ################################################################################
-# oops/inst/juno/__init__.py
+# oops/inst/juno/juno_.py
```
rfrenchseti, line 2: Wrong filename?

PR #15, opened 2022-09-27
Unittests

jnspitale:
The purpose here is to create a stand-alone exercise_backplanes module that can be imported and used for unit tests by any host.


oops/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
rfrenchseti, line 1: Please don't checkin your local .vscode settings because not everyone wants to use the same settings. I thought I had changed .gitignore to prevent this but apparently what I did wasn't sufficient.

markshowalter (approved):
Joe--These changes look fine. Give a moment's thought to my "lrange" recommendation but you can go either way on that. I agree that it makes sense to reorganize the backplane unit tests as you have done, but I'm sure you realize this is just a step along the way. These unit tests are still hard-wired to Saturn, the rings, and Epimetheus, so making allowances for different instruments is only a partial solution. Going forward, maybe a way to think about this is to break the existing "exercises" down into smaller ones, in which the planet, ring(s) and moon(s) are input parameters. Then the full suite of unit tests for all our backplanes and also for all our host/instrument classes would involve running a pre-defined set of exercises against a curated list of observations, each with an appropriate set of planets, rings, and moons.

rfrenchseti (changes requested):
Marking this officially as "changes requested" because of the .vscode directory

jnspitale:
Yes, I do realize that and I'm looking breaking it down like that now.  I just wanted to indicate the general direction I wanted to go before spending a lot of time going down a dead-end road.

-joe


________________________________

jnspitale:
Regarding lrange, I am in agreement.  I suspect what happened is that I initially changed it to list(range) after noticing that range() no longer returns a list in pyhton3.  You probably subsequently defined lrange and when I saw that code I misread it, and thought my original change had not been committed.  It looks good now.

-joe



________________________________

PR #19, opened 2022-11-16
Unittests2

jnspitale:
Reference backplanes have already been generated and reside in dropbox, so the unit tests should find them and run properly.

hosts/ has its own master unit tester that is not referenced by the oops unit tester because they are parallel in the tree, and oops is not supposed to know specifics of the hosts.


oops/backplane/distance.py
```diff
@@ -140,3 +140,292 @@ def center_time(self, event_key):
...
+        test = bp.distance(planet+':limb')
+        show_info(bp, 'Distance observer to planet limb (km)', test,
+                  printing=printing, saving=saving, dir=dir, refdir=refdir)
+
+        ### TODO: use Body.has_rings() once Mark has created it
+        try:
+            test = bp.distance(planet+':ansa')
+            show_info(bp, 'Distance observer to ansa (km)', test,
+                      printing=printing, saving=saving, dir=dir, refdir=refdir)
+        except:
```
rfrenchseti, line 186: In general we don't want a naked `except`. It would be nice to know what you're looking for here.

oops/backplane/distance.py
```diff
@@ -140,3 +140,292 @@ def center_time(self, event_key):
...
+from oops.backplane.unittester_support  import show_info
+
+
+#===========================================================================
+def exercise_observer(bp, printing, saving, dir, refdir,
+                      planet=None, moon=None, ring=None,
+                      undersample=16, use_inventory=False, inventory_border=2):
+    """generic unit tests for distance.py"""
+
+    if planet != None:
```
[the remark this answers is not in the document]
rfrenchseti: Same for many other tests below I won't call out individually

oops/backplane/distance.py
```diff
@@ -140,3 +140,292 @@ def center_time(self, event_key):
...
+                  printing=printing, saving=saving, dir=dir, refdir=refdir)
+
+        test = bp.center_distance(ring)
+        show_info(bp, 'Distance observer to ring center (km)', test,
+                  printing=printing, saving=saving, dir=dir, refdir=refdir)
+
+    if moon != None:
+        test = bp.distance(moon)
+        show_info(bp, 'Distance observer to moon (km)', test,
+                     printing=printing, saving=saving, dir=dir, refdir=refdir)
```
rfrenchseti, line 201: Bad indent

oops/backplane/exercise_backplanes.py
```diff
@@ -144,872 +34,224 @@ def show_info(title, array):
...
-    test = bp.sub_solar_latitude('saturn')
-    show_info('Saturn sub-solar latitude (deg)', test * DPR)
-
-    test = bp.latitude('epimetheus', lat_type='centric')
-    show_info('Epimetheus latitude, planetocentric (deg)', test * DPR)
+    if printing:
+        print('\n********* right ascension')
+    import oops.backplane.sky as sky
+    sky.exercise_right_ascension(bp, printing, saving, dir, refdir,
+                               planet=planet_key, moon=moon_key, ring=ring_key)
```
rfrenchseti, line 43: Bad indents here and below

oops/backplane/limb.py
```diff
@@ -83,3 +83,46 @@ def _fill_limb_intercepts(self, event_key, limit=None, lock_limits=False):
...
+
+
+
+
+################################################################################
+# UNIT TESTS
+################################################################################
+
+import unittest
+from oops.meshgrid     import Meshgrid
```
rfrenchseti, line 95: Spacing is messy and inconsistent

oops/backplane/ring.py
```diff
@@ -829,3 +829,312 @@ def ring_radius_in_front(self, event_key, ring_body):
...
+#===========================================================================
+def exercise_radial_modes(bp, printing, saving, dir, refdir,
+                          planet=None, moon=None, ring=None,
+                          undersample=16, use_inventory=False, inventory_border=2):
+    """Generic unit tests for ring.py"""
+
+    if ring != None:
+        test = bp.ring_radius(ring)
+        show_info(bp, 'Ring radius (km)', test,
+                     printing=printing, saving=saving, dir=dir, refdir=refdir)
```
rfrenchseti, line 877: Bad indent here and below

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+                              the type, e.g., '.png' or '.jpg'
+    """
+    try:
+        im = Image.open(filename, mode="r")
+    except:
+        return None
+
+    return Scalar.as_scalar((np.array(im)))
+
+
```
rfrenchseti, line 30: Mark should chime in here, but in general it seems like the amount of vertical whitespace is getting excessive everywhere. One of our coding style guidelines is to make as much fit on the screen vertically at a time as possible. I don't think having 3 or 4 blank lines PLUS a divider is necessary for readability.

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+def _read_image(filename):
+    """
+    Save an image file of a 2-D array.
+
+    Input:        filename    the name of the input file, which should end with
+                              the type, e.g., '.png' or '.jpg'
+    """
+    try:
+        im = Image.open(filename, mode="r")
+    except:
```
rfrenchseti, line 25: Bad naked except

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+    return filename.lower().rstrip('_')
+
+
+
+#===============================================================================
+def _construct_filename(bp, array, title, dir):
+    """Constructs a backplane filename."""
+
+    # Construct base filename from title
+    filename = _convert_filename('backplane-' + \
```
rfrenchseti, line 113: \ is not needed

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+    # NOTE: if no backplane key exists, a non-unique filename could result
+    try:
+        key = array.key
+        id = key[0]
+        for item in key[1:]:
+            if item != ():
+                id = id + '_' + str(item)
+        id = _convert_filename(id)
+        filename = filename + '[' + id + ']' + '.png'
+    except:
```
rfrenchseti, line 126: Bad naked except. Also too much code inside the try. Lots of chance for false exceptions.

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+
+    _print(title, printing=printing)
+
+
+    # Scalar summary
+    if isinstance(array, numbers.Number):
+        _print('  ', array, printing=printing)
+
+    # Mask summary
+    elif type(array.vals) == bool or \
```
rfrenchseti, line 167: use () instead of \

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+            (isinstance(array.vals, np.ndarray) and \
+             array.vals.dtype == np.dtype('bool')):
+        count = np.sum(array.vals)
+        total = np.size(array.vals)
+        percent = int(count / float(total) * 100. + 0.5)
+        _print('  ', (count, total-count),
+            (percent, 100-percent), '(True, False pixels)', printing=printing)
```
rfrenchseti, line 174: Bad indent

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+        maxval = np.max(array.vals)
+        if minval == maxval:
+            _print('  ', minval, printing=printing)
+        else:
+            _print('  ', (minval, maxval), '(min, max)', printing=printing)
+
+    # Masked backplane summary
+    else:
+        _print('  ', (array.min().as_builtin(),
+            array.max().as_builtin()), '(masked min, max)', printing=printing)
```
rfrenchseti, line 190: Bad indent

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+        total = np.size(array.mask)
+        masked = np.sum(array.mask)
+        percent = int(masked / float(total) * 100. + 0.5)
+        _print('  ', (masked, total-masked),
+             (percent, 100-percent), '(masked, unmasked pixels)', printing=printing)
```
rfrenchseti, line 195: Bad indent

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+    import sys
+
+
+    ## Define arguments ##
+    parser = argparse.ArgumentParser(description='Backplane unit tester.')
+
+
+    # Generic arguments
+    parser.add_argument('--args', nargs='*', metavar='arg', default=None,
+                        help='Generic arguments to pass to the test modules. ' \
```
rfrenchseti, line 381: \ not needed

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+                        help='Execute only the backplane exercises.')
+
+    parser.add_argument('--no-exercises', action='store_true', default=None,
+                        help='Execute all tests except the backplane exercises.')
+
+    parser.add_argument('--no-compare', action='store_true', default=None,
+                        help='Do not compare backplanes with references.')
+
+    parser.add_argument('--output', nargs=1, metavar='dir', default=None,
+                        help='Directory in which to save backplane PNG images. ' \
```
rfrenchseti, line 402: \ not needed

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+
+
+    # Generic arguments
+    parser.add_argument('--args', nargs='*', metavar='arg', default=None,
+                        help='Generic arguments to pass to the test modules. ' \
+                             'Must occur last in the argument list.')
+
+
+    # Basic controls
+    parser.add_argument('--verbose', action='store_true', default=None,
```
rfrenchseti, line 386: I think the default for a boolean flag should be False instead of None

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+class Backplane_Settings(object):
+
+    from oops.unittester_support            import TESTDATA_PARENT_DIRECTORY
+
+    ARGS = []
+    DIFF = []
+    EXERCISES_ONLY = False
+    NO_EXERCISES = False
+    NO_COMPARE = False
+    SAVING = True
```
rfrenchseti, line 351: I don't think the default should be to save the backplane files. Perhaps this is left over from debugging?

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+        os.makedirs(dir, exist_ok=True)
+
+        filename = _construct_filename(bp, array, title, dir)
+        _save_image(image, filename)
+
+
+    # Compare with reference array if refdir is known
+    if refdir is not None:
+        assert os.path.exists(refdir), \
+            "No reference directory.  Use --no-compare, --no-exercises, or --reference."
```
rfrenchseti, line 228: It would be helpful to add the reference directory to the error message. I hit this assert but had no idea what it was looking for.

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,503 @@
...
+    """Internal method to print summary information and display images as
+    desired.
+    """
+
+    import numbers
+
+    if array is None:
+        return
+
+    if not printing and not saving and refdir==None:
```
rfrenchseti, line 156: is None

oops/backplane/exercise_backplanes.py
```diff
@@ -144,872 +34,224 @@ def show_info(title, array):
...
+#===============================================================================
+def exercise_backplanes_settings(obs):
+    """Wrapper for exercise_backplanes(). Determines default directories
+       based on input observation object and fills in args based on
+       settings."""
 
-    test = bp.border_atop(('ring_radius', 'saturn:ring'), 100.e3)
-    show_info('Border of ring radius atop 100 kkm', test)
+    # determine default output and reference directories
+    specname = obs.filespec.split(TESTDATA_PARENT_DIRECTORY)[1]
```
rfrenchseti, line 219: This doesn't work for me and causes all of the backplane tests to fail. `obs.filespec` is something like `/test_dir/cassini/ISS/file.IMG` then `specname` is `/cassini/ISS/file.IMG` then `basedir` is `/cassini/ISS`. Then the creation of `outdir` thinks that `basedir` is an absolute path so it ignores all of the items in the path before than. Is it possible you're making your TESTDATA_PARENT_DIRECTORY with a slash at the end? This should not be necessary and the code should handle both cases.

rfrenchseti (changes requested):
- When a formatting issue occurred in multiple files I only commented on it the first time or two
- I can't run the Juno host unittests because they refer to kernels in `/home/jspitale`
- I'll do another review after these changes are made

hosts/juno/jiram/img.py
```diff
@@ -266,3 +264,62 @@ def reset():
...
+
+        if Backplane_Settings.NO_EXERCISES:
+            self.skipTest("")
+
+        root = os.path.join(TESTDATA_PARENT_DIRECTORY, "juno/jiram")
+
+
+        # Moon image
+        file = os.path.join(root, "JNOJIR_2000/DATA/JIR_IMG_RDR_2013282T133843_V03.IMG")
+        _obs = jiram.from_file(file); body_name = "MOON"; obs = _obs[1]
```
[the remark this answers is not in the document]
jnspitale: This carried over from my test scripts where it was more conventient to write it this way for the way I was using them.

oops/backplane/unittester_support.py
jnspitale: This is intentional.  I'm using the defaults defined as Backplane_Settings attributes, so  defaulting everything to None here allows me to detect whether or not the argument has been supplied, and therefore whether to change it from the what was set there.  I find this much easier to read and manage.

oops/backplane/unittester_support.py
jnspitale: This was just my preference.  If a test fails, the output files will already exist and can be inspected without having to re-run the tests with a different argument list.

oops/backplane/unittester_support.py
jnspitale: Agreed, I meant to do that.

hosts/juno/jiram/__init__.py
```diff
@@ -52,16 +50,16 @@ def from_file(filespec,
     # Image
     if ext.upper() == '.IMG':
         from . import img
-        return(img.from_file(filespec, label,
-                                     return_all_planets=False, **parameters))
+        return img.from_file(filespec, label,
+                                     return_all_planets=False, **parameters)
```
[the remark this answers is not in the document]
jnspitale: Still catching old indent issues.  I am using python indent now, so this should be less of an issue going forward.

hosts/juno/jiram/img.py
```diff
@@ -211,8 +209,8 @@ class IMG(object):
     #===========================================================================
     @staticmethod
     def initialize(time, ck='reconstructed', planets=None, asof=None,
-               spk='reconstructed', gapfill=True,
-               mst_pck=True, irregulars=True):
+                         spk='reconstructed', gapfill=True,
```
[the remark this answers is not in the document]
jnspitale: Yeah, this is definetly not pep-8 compliant.  It's consistent with a lot of existing oops code and I find it more readable.

oops/backplane/exercise_backplanes.py
jnspitale: Yes, I have my $OOPS_TEST_DATA_PATH defined with a trailing /, so this make sense.  I'll use os.path.dirname() to standardize TESTDATA_PARENT_DIRECTORY before splitting the path.

oops/backplane/unittester_support.py
rfrenchseti: OK. This was causing me problems since save was crashing, but hopefully that won't be a problem going forward.

hosts/juno/jiram/img.py
```diff
@@ -4,22 +4,20 @@
...
 import julian
-import pdstable
 import cspyce
 from polymath import *
 import os.path
-import pdsparser
 import oops
 
-from . import JIRAM
+from hosts.juno.jiram import JIRAM
```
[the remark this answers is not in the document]
jnspitale: The problem is with '.' you can't run this module's main() using:
% python img.py

Instead you have to use:
% python -m hosts.juno.jiram.img

This is a bit more of a pain to deal with.  Also (though perhaps not relevant in this case), it wouldn't be interpreted correctly in a shebang because the -m would get bundled with the command instead of being an argument to python.  Also, pep-8 discourages relative imports, though it does allow for the form you mention above: '.juno.jiram'

oops/backplane/__init__.py
```diff
@@ -7,11 +7,11 @@
 import numpy as np
 from polymath import Boolean, Qube, Scalar
 
-from ..body              import Body
-from ..path              import AliasPath
-from ..surface.ansa      import Ansa
-from ..surface.limb      import Limb
-from ..surface.ringplane import RingPlane
+from oops.body              import Body
```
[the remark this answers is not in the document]
jnspitale: see above; Mark and I discussed relative imports and decided to switch back to absolute imports.

oops/backplane/ansa.py
```diff
@@ -284,3 +284,100 @@ def ansa_vertical_resolution(self, event_key):
...
+import unittest
+from oops.meshgrid                       import Meshgrid
+from oops.unittester_support             import TESTDATA_PARENT_DIRECTORY
+from oops.constants                      import DPR
+from oops.backplane.unittester_support   import show_info
+
+
+#===========================================================================
+def exercise_resolution(bp, printing, saving, dir, refdir,
+                        planet=None, moon=None, ring=None,
```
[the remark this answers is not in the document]
jnspitale: That's because there are no optional arguments on the first line to line up with.  My preferred indentation actually would have been line up planet and printing, as bp has particular 
importance.

oops/backplane/exercise_backplanes.py
```diff
@@ -144,872 +33,220 @@ def show_info(title, array):
...
-    test = bp.where_below(('ring_radius', 'saturn_main_rings'), 10.e3)
-    show_info('Empty mask of Saturn ring radius below 10 kkm', test)
+#===============================================================================
+def exercise_backplanes_settings(obs):
+    """Wrapper for exercise_backplanes(). Determines default directories
+       based on input observation object and fills in args based on
+       settings."""
+
+    # determine default output and reference directories
+    specname = obs.filespec.split( \
```
rfrenchseti, line 215: This still doesn't work if the TESTDATA_PARENT_DIRECTORY doesn't end with a /.

oops/backplane/distance.py
rfrenchseti: Not resolved. If the tests don't work, perhaps just comment them out? There are similar examples in several other backplane files.

oops/backplane/distance.py
```diff
@@ -140,3 +140,276 @@ def center_time(self, event_key):
...
+        show_info(bp, 'Distance Sun to ring center, arrival (km)', test,
+                  printing=printing, saving=saving, dir=dir, refdir=refdir)
+
+        test = bp.center_distance(('sun', ring), direction='dep')
+        show_info(bp, 'Distance Sun to ring center, departure (km)', test,
+                  printing=printing, saving=saving, dir=dir, refdir=refdir)
+
+#===============================================================================
+def exercise_observer_light_time(bp, printing, saving, dir, refdir,
+                      planet=None, moon=None, ring=None,
```
rfrenchseti, line 258: Bad indent

oops/backplane/exercise_backplanes.py
rfrenchseti: Not resolved

oops/backplane/resolution.py
```diff
@@ -120,3 +120,178 @@ def _fill_surface_resolution(self, event_key):
...
+                  printing=printing, saving=saving, dir=dir, refdir=refdir)
+
+        test = bp.coarsest_resolution(moon)
+        show_info(bp, 'moon coarsest resolution (km)', test,
+                  printing=printing, saving=saving, dir=dir, refdir=refdir)
+
+    if ring is not None:
+        test = bp.resolution(ring, 'u')
+        show_info(bp, 'Ring resolution along u axis (km)', test,
+                     printing=printing, saving=saving, dir=dir, refdir=refdir)
```
rfrenchseti, line 194: Bad indent

oops/backplane/ring.py
```diff
@@ -829,3 +829,301 @@ def ring_radius_in_front(self, event_key, ring_body):
...
 
 ################################################################################
+
+
+################################################################################
+# UNIT TESTS
+################################################################################
+
+import unittest
+from oops.meshgrid     import Meshgrid
```
rfrenchseti, line 839: Other files have these lined up

oops/backplane/unittester.py
```diff
@@ -1,289 +1,58 @@
...
+#  -h, --help            show this help message and exit
+#  --args [arg [arg ...]]
+#                        Generic arguments to pass to the test modules. Must
+#                        occur last in the argument list.
+#  --verbose             Print output to the terminal.
+#  --diff old new        Compare new and old backplane logs.
+#  --exercises-only      Execute only the backplane exercises.
+#  --no-exercises        Execute all tests except the backplane exercises.
+#  --no-compare          Do not compare backplanes with references.
+#  --output dir          Directory in which to save backplane PNG images.
```
rfrenchseti, line 20: So does this mean that every time someone runs the tests, it regenerates the PNG files and updates the files in Dropbox? That seems bad. The work done by a developer should stay local to their workstation and only update the global dropbox when specifically requested to. In fact if a person has read-only access to the folder, like would happen on an test server, the directory won't even be writeable. Perhaps there needs to be a different default destination directory.

oops/backplane/unittester.py
```diff
@@ -1,289 +1,58 @@
...
+#                        Default is TESTDATA_PARENT_DIRECTORY/[data dir]. If
+#                        the directory does not exist, it is created.
+#  --no-output           Disable saving of backplane PNG files.
+#  --log                 Enable the internal oops logging.
+#  --undersample N       Amount by which to undersample backplanes. Default is
+#                        16.
+#  --reference           Generate reference backplanes and exit.
+#  --test-level N        Selects among pre-set parameter combinations:
+#                        -test_level 1: no printing, no saving, undersample 32.
```
rfrenchseti, line 29: The help message has the wrong string for the argument.

oops/backplane/unittester_support.py
rfrenchseti: Not resolved

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,483 @@
...
+                        help='Amount by which to undersample backplanes.  Default is 16.')
+
+
+    parser.add_argument('--reference', action='store_true', default=None,
+                        help='Generate reference backplanes and exit.')
+
+#TODO: currently only works for 80-char width
+    parser.add_argument('--test-level', nargs=1, type=int, metavar='N', default=None,
+                        help='Selects among pre-set parameter combinations:  '
+                              '-test_level 1: no printing, no saving, undersample 32. '
```
rfrenchseti, line 402: Fix argument strings in help message

rfrenchseti (changes requested):
- I would still like Mark to chime in on the amount of vertical whitespace, sometimes as many as four blank lines in a row.

oops/backplane/distance.py
jnspitale: Ok, I commented them out and left the warning message.

hosts/juno/junocam.py
```diff
@@ -61,23 +60,25 @@ def from_file(filespec, fast_distortion=True,
...
-                                 instrument = "JUNOCAM",
-                                 filter = fmeta.filter,
-                                 data = framelets[:,:,i]))
+            item = (oops.obs.Snapshot(
+                ("v","u"),
+                (fmeta.tstart, fmeta.tdi_texp), fmeta.fov,
+                "JUNO", "JUNO_JUNOCAM",
+                instrument = "JUNOCAM",
+                filter = fmeta.filter,
+                data = framelets[:,:,i]))
```
markshowalter, line 69: This indentation is very hard to read. It wasn't obvious that this was a Snapshot call and not just a very long tuple. The arguments to Snapshot should all be indented at least as far as the first space after "Snapshot(" in the first line.

Also, we have been generally using single quotes around strings in oops, rather than double quotes. I do sometimes find some of my old code that used double quotes. I strongly recommend that we standardize on single quotes for all new code. Elsewhere, I tend to leave existing files alone, but I convert double quotes to single when I have some other reason to be working on a file.

markshowalter (changes requested):
My comments are mostly stylistic. I hope a lot of them can be implemented through global search and replace.

I have not tried to run these changes, but it appears Rob is working with you on that. I leave it to you and Rob to figure out how best to implement the tests so that they don't normally overwrite what's on Dropbox.

hosts/juno/jiram/img.py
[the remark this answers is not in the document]
markshowalter: > We generally discourage the use of `;` and multiple statements on one line. Perhaps Mark can chime in on this.

I agree. I don't think I have ever used a semicolon in my code. They can be handy sometimes for interactive Python.

oops/backplane/exercise_backplanes.py
markshowalter: The way I handle optional trailing slashes in directory paths is dir.rstrip('/') if the trailing slash should not be there and dir.rstrip('/') + "/" if it is required. Of course, os.path.join is the proper way to handle this, but sometimes I find that too cumbersome.

By the way, you never need a backslash after an opening paren.

oops/backplane/spheroid.py
```diff
@@ -6,16 +6,16 @@
...
-
-############################################################################
+################################################################################
 # Body surface geometry, surface intercept versions
 #   longitude()
 #   latitude()
-############################################################################
+################################################################################
 
+#===============================================================================
```
markshowalter, line 18: Avoid one horizontal separator immediately after another. It seems unnecessary.

oops/backplane/unittester_backplanes_log_diff.py
```diff
@@ -1,112 +0,0 @@
-#!/usr/bin/env python3
```
markshowalter, line 1: I am quite happy to see this log differencer go away. However, rather than delete it entirely, please just move it to the ideas subdirectory. I would not want to start writing it again from scratch if we ever need something like it in the future.

oops/backplane/where.py
```diff
@@ -205,3 +204,114 @@ def where_all(self, *backplane_keys):
...
+
+#===========================================================================
+def exercise(bp, printing, saving, dir, refdir,
+             planet=None, moon=None, ring=None,
+             undersample=16, use_inventory=False, inventory_border=2):
+    """generic unit tests for where.py"""
+
+    if planet is not None:
+        test = bp.where_intercepted(planet)
+        show_info(bp, 'Mask of planet intercepted', test,
```
markshowalter, line 226: By the way, because you are using the same last four arguments in every call to show_info, I suggest you create a dictionary called "options", with these four items, and then change the call to show_info(..., **options). That will also make it easier to add some new option in the future if we decide we need it.

markshowalter (changes requested):
See comments from my first review.

oops/backplane/spheroid.py
jnspitale: Ok I would prpose that the closing ####### separator of the preceding comment block should be removed (here and in general) so that method definitions have a consistent look.

oops/backplane/exercise_backplanes.py
jnspitale: Ok, I see the same misbehavior as Rob using his definition of $OOPS_TEST_DATA_PATH, so I just used os.path.normpath() to standardize TESTDATA_PARENT_DIRECTORY where it is defined in oops/unitester_support, and included the trailing slash.  All other references to TESTDATA_PARENT_DIRECTORY are as an argument to os.path.join() so they don't care.

oops/backplane/spheroid.py
rfrenchseti: I would think the big text block should have `####` before and after, so that it looks like a consistent block (3-sided rectangle). I see no reason to put any `#====` before a function definition. That is not something we do in any other code base.

jnspitale:
That's why I find most code virtually unreadble.


________________________________

hosts/cassini/iss.py
```diff
@@ -138,6 +140,8 @@ def from_index(filespec, **parameters):
 
     return snapshots
 
+
+
```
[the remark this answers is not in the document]
jnspitale: That is consistent with pep-8.  I think most of Mark's consectuve blank lines occur between the main code and the unit tests, which is also consistent with pep-8's exception for separating logical code units.

hosts/voyager/iss.py
```diff
@@ -391,3 +391,42 @@ def initialize(asof=None):
 ISS.initialize()
 
 ################################################################################
+
+
+
+################################################################################
```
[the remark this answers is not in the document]
jnspitale: I think horizontal separators are useful, so I'd like to come to a consensus on how to use them.  Generally, within the bulk of a file, Mark uses #### and I use ==== and **** for functions and classes.  I don't care what we use, but I'd like to standardize this so I can stop guessing / imposing my own conventions.  Rob has suggested not delineating methods, but what about classes?

hosts/unittester.py
```diff
@@ -4,11 +4,17 @@
...
 
-from juno.junocam   import Test_Juno_Junocam
-#from cassini.iss    import Test_Cassini_ISS
-#from hst            import Test_HST
+from hosts.juno.unittester            import *
+#from hosts.cassini.unittester         import *
+#from hosts.hst.unittester             import *
+#from hosts.voyager.unittester         import *
+#from hosts.newhorizons.unittester     import *
+#from hosts.keck.unittester            import *
```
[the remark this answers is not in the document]
jnspitale: Do we care what flake8 thinks of the unit tests?  In this case, * seems to make sense because we actually intend to run every test contained in the module.  Without *, we have to explicitly add any new tests to this file when we add them to the referenced module.  I don't see the sense in that.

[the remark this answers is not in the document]
rfrenchseti: I agree. I think this is one of the rare cases where import * makes perfect sense.

hosts/voyager/iss.py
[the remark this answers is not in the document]
rfrenchseti: I will let Joe and Mark decide on what they want for this code base. I notice that Mark does put a horizontal divider before every class method in polymath and oops. I personally don't use any (I think between color coding and obvious doc strings the start of a method is easy to see) but I don't need to force that on others if the consensus is they are useful.

oops/backplane/exercise_backplanes.py
jnspitale: this should be fixed now, so go ahead and try it and let me know

rfrenchseti:
OOPS backplane tests now work without a trailing space on the TEST_DIR variable and backplane results are stored in a separate directory.

rfrenchseti:
polymath, oops, and hosts tests all pass

rfrenchseti:
The change to spicedb.py does not work. At the point where you put the test for file existence, the `kernel.filespec` can actually be a list of kernels, like this:

```
Uranus/SPK/old/ura083.bsp,Voyager/SPK/vgr2_ura083.bsp
```

Instead you need to undo the changes you made, and put the following further down the function:

```     for abspath in abspaths:

            # Remove the name from earlier in the list if necessary
            if abspath in abspath_list:
                abspath_list.remove(abspath)

            # Always add it at the end
            abspath_list.append(abspath)
            abspath_types[abspath] = kernel.kernel_type     # track kernel types

            if not os.path.exists(abspath):
                warnings.warn('SPICE kernel not found: ' + testpath, RuntimeWarning)
            else:
                # Save the info for each furnished file
                basename = os.path.basename(abspath)
                if basename in FURNISHED_INFO:
                    if kernel not in FURNISHED_INFO[basename]:
                        FURNISHED_INFO[basename].append(kernel)
                else:
                    FURNISHED_INFO[basename] = [kernel]
```

rfrenchseti:
Sorry that should be `warnings.warn('SPICE kernel not found: ' + abspath, RuntimeWarning)`

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,389 @@
...
+                              '-test_level 2: printing, no saving, undersample 16. '
+                              '-test_level 3: printing, saving, no undersampling. '
+                              'These behaviors are overridden by other arguments.')
+
+
+    ## Parse arguments, leaving unknown args for some other parser ##
+    argv, left = parser.parse_known_args()
```
rfrenchseti, line 313: It's confusing that this variable is called `argv` when it isn't actually the `argv` or any part of it, it's the result of the argument parsing. Perhaps rename to something like `args`.

rfrenchseti (changes requested):
All hosts, polymath, and oops tests pass. Most command line arguments tested. Just a couple of minor issues remain (not related to coding style).

oops/backplane/unittester_support.py
jnspitale: Changed ARGS to GENERIC_ARGS and ARGV to ARGS.

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,389 @@
...
+def backplane_unittester_args():
+    """Parse command-line arguments for backplane unit tests.
+
+    This function must be called by main() in order for these arguments to
+    to be parsed.  If not, the above default values will apply."""
+    import argparse
+    import sys
+
+    ## Define arguments ##
+    parser = argparse.ArgumentParser(description='Backplane unit tester.')
```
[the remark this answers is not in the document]
jnspitale: I haven't been able to get it to do that today, but I'm still playing with it.  I have a note to try to figure this out, but perhaps we should move forward with the merge and I'll keep working on this.  If all else fails I can revert to the earlier direct parsing of sys.argv.

hosts/juno/junocam.py
markshowalter: I thought this was going to be fixed. Please fix.

hosts/juno/junocam.py
```diff
@@ -61,23 +60,25 @@ def from_file(filespec, fast_distortion=True,
...
-                                 filter = fmeta.filter,
-                                 data = framelets[:,:,i]))
+            item = (oops.obs.Pushframe(
+                        ("vt","u"),
+                        (fmeta.tstart, fmeta.tdi_texp, fmeta.tdi_stages),
+                        fmeta.fov,
+                        "JUNO", "JUNO_JUNOCAM",
+                        instrument = "JUNOCAM",
+                        filter = fmeta.filter,
+                        data = framelets[:,:,i]))
```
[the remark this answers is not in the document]
markshowalter: Again, please fix the indentation. Also, the outer parentheses surrounding the entire expression appear to be superfluous. Please remove.

hosts/voyager/iss.py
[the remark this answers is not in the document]
markshowalter: I'm confused, because I am not seeing any changes since last time. There should never be three blank lines in a row anywhere. There should only be two (at most) before classes and some top-level function defs. There should never be two horizontal lines with only blank lines between them.

oops/backplane/border.py
```diff
@@ -137,3 +136,32 @@ def border_outside(self, backplane_key):
 Backplane._define_backplane_names(globals().copy())
 
 ################################################################################
+
+
+
+################################################################################
```
markshowalter, line 142: remove lines 138-141 inclusive.

oops/backplane/border.py
```diff
@@ -137,3 +136,32 @@ def border_outside(self, backplane_key):
...
+from oops.constants                    import DPR
+from oops.backplane.unittester_support import show_info
+
+
+#*******************************************************************************
+class Test_Border(unittest.TestCase):
+
+    #===========================================================================
+
+    def runTest(self):
```
markshowalter, line 157: There should never be a blank line between a horizontal separator and a def. Remove.

hosts/cassini/iss.py
```diff
@@ -415,27 +414,58 @@ def reset():
 
         Cassini.reset()
 
+
+
```
markshowalter, line 418: Why was this changed from one blank line (correct) to three blank lines (incorrect). Please fix.

hosts/juno/jiram/img.py
```diff
@@ -47,12 +44,12 @@ def from_file(filespec, label, fast_distortion=True,
...
-                                 "JUNO", "JUNO_JIRAM_I_" + fmeta.filter_frame,
-                                 instrument = "JIRAM_I",
-                                 filter = fmeta.filter,
-                                 data = framelets[:,:,i]))
+        item = (oops.obs.Snapshot(('v','u'),
+                                  fmeta.tstart, fmeta.exposure, fmeta.fov,
+                                  'JUNO', 'JUNO_JIRAM_I_' + fmeta.filter_frame,
+                                  instrument = 'JIRAM_I',
+                                  filter = fmeta.filter,
+                                  data = framelets[:,:,i]))
```
markshowalter, line 52: Remove the surrounding parentheses. They are not needed.

hosts/juno/jiram/spe.py
```diff
@@ -199,14 +192,50 @@ def initialize(time, ck='reconstructed', planets=None, asof=None,
...
+
+        Can be useful for debugging.
         """
         SPE.initialized = False
 
         JIRAM.reset()
 
 
 
+################################################################################
```
markshowalter, line 205: Change to one blank line.

hosts/juno/jiram/spe.py
```diff
@@ -199,14 +192,50 @@ def initialize(time, ck='reconstructed', planets=None, asof=None,
...
+
+        root = os.path.join(TESTDATA_PARENT_DIRECTORY, 'juno/jiram')
+        file = os.path.join(root, 'JNOJIR_2000/DATA/JIR_SPE_RDR_2013282T133845_V03.DAT')
+        (obs, slits) = jiram.from_file(file)
+        exercise_backplanes(obs, use_inventory=True, inventory_border=4,
+                                 planet_key='MOON')
+
+
+
+##############################################
```
markshowalter, line 235: Remove two of the three blank lines here.

hosts/juno/junocam.py
```diff
@@ -52,51 +50,43 @@ def from_file(filespec, fast_distortion=True,
...
-                                 data = framelets[:,:,i]))
-
-
+            item = (oops.obs.Snapshot(
+                                ('v','u'),
+                                (fmeta.tstart, fmeta.tdi_texp), fmeta.fov,
+                                'JUNO', 'JUNO_JUNOCAM',
+                                instrument = 'JUNOCAM',
+                                filter = fmeta.filter,
+                                data = framelets[:,:,i]))
```
markshowalter, line 65: Remove the surrounding parentheses.

hosts/juno/junocam.py
```diff
@@ -52,51 +50,43 @@ def from_file(filespec, fast_distortion=True,
...
-                                 data = framelets[:,:,i]))
-
+            item = (oops.obs.Pushframe(
+                                    ('vt','u'),
+                                    (fmeta.tstart, fmeta.tdi_texp, fmeta.tdi_stages),
+                                    fmeta.fov,
+                                    'JUNO', 'JUNO_JUNOCAM',
+                                    instrument = 'JUNOCAM',
+                                    filter = fmeta.filter,
+                                    data = framelets[:,:,i]))
```
markshowalter, line 75: ditto

oops/backplane/distance.py
```diff
@@ -140,3 +140,197 @@ def center_time(self, event_key):
 Backplane._define_backplane_names(globals().copy())
 
 ################################################################################
+
+
+################################################################################
+# UNIT TESTS
```
markshowalter, line 146: delete lines 143-145

oops/backplane/orbit.py
```diff
@@ -87,3 +87,50 @@ def orbit_longitude(self, event_key, reference='obs', planet=None):
 Backplane._define_backplane_names(globals().copy())
 
 ################################################################################
+
+
+################################################################################
```
markshowalter, line 92: delete lines 90-92

oops/backplane/pole.py
```diff
@@ -67,3 +67,43 @@ def pole_position_angle(self, event_key):
 Backplane._define_backplane_names(globals().copy())
 
 ################################################################################
+
+
+################################################################################
```
markshowalter, line 72: delete lines 70-72

oops/backplane/resolution.py
```diff
@@ -120,3 +119,124 @@ def _fill_surface_resolution(self, event_key):
 Backplane._define_backplane_names(globals().copy())
 
 ################################################################################
+
+
+################################################################################
```
markshowalter, line 124: delete lines 122-124

oops/backplane/ring.py
```diff
@@ -829,3 +829,207 @@ def ring_radius_in_front(self, event_key, ring_body):
 Backplane._define_backplane_names(globals().copy())
 
 ################################################################################
+
+
+################################################################################
```
markshowalter, line 834: delete lines 832-834

oops/backplane/sky.py
```diff
@@ -244,3 +243,91 @@ def _fill_center_ra_dec(self, event_key, apparent, direction):
 Backplane._define_backplane_names(globals().copy())
 
 ################################################################################
+
+
+################################################################################
```
markshowalter, line 248: delete lines 246-248

oops/backplane/spheroid.py
```diff
@@ -482,3 +482,169 @@ def lommel_seeliger_law(self, event_key):
 Backplane._define_backplane_names(globals().copy())
 
 ################################################################################
+
+
+################################################################################
```
markshowalter, line 487: delete lines 485-487

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,389 @@
...
+
+        if total == masked:
+            minval = np.min(array.vals)
+            maxval = np.max(array.vals)
+        else:
+            minval = array.min().as_builtin()
+            maxval = array.max().as_builtin()
+
+
+    # The rest of the method applies only to arrays
```
markshowalter, line 192: never two blank lines in a row in the middle of a function

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,389 @@
...
+    if array.shape == ():
+        return
+
+
+    # scale image
```
markshowalter, line 197: ditto. And a few more occurrences below.

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,389 @@
...
+    REF = False
+
+    # TBD attributes
+#    DIFF = []
+#    PLANET_KEY = None
+#    MOON_KEY = None
+#    RING_KEY = None
+
+
+    # Attributes that depend on the input file.  Defaults are determined
```
markshowalter, line 245: ditto

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,389 @@
...
+    parser = argparse.ArgumentParser(description='Backplane unit tester.')
+
+
+    # Generic arguments
+    parser.add_argument('--args', nargs='*', metavar='arg', default=None,
+                        help="""Generic arguments to pass to the test modules.
+                                Must occur last in the argument list.""")
+
+
+    # Basic controls
```
markshowalter, line 270: ditto

oops/backplane/unittester_support.py
```diff
@@ -0,0 +1,389 @@
...
+        Backplane_Settings.UNDERSAMPLE = args.undersample[0]
+
+    if args.reference is not None:
+        Backplane_Settings.EXERCISES_ONLY = True
+        Backplane_Settings.NO_COMPARE = True
+        Backplane_Settings.SAVING = True
+        Backplane_Settings.REF = True
+
+
+#   TBD
```
markshowalter, line 374: ditto

oops/backplane/where.py
```diff
@@ -205,3 +204,80 @@ def where_all(self, *backplane_keys):
 Backplane._define_backplane_names(globals().copy())
 
 ################################################################################
+
+
+################################################################################
```
markshowalter, line 209: delete 207-209

markshowalter (changes requested):
I'm not sure what happened but I don't see changes in response to most of my previous comments. I have augmented the list of comments this time. Please fix.

oops/backplane/border.py
jnspitale: Addressed these and elsewhere

oops/backplane/distance.py
jnspitale: Addressed here and elsewhere

oops/backplane/ansa.py
```diff
@@ -284,3 +284,93 @@ def ansa_vertical_resolution(self, event_key):
...
+                      printing=printing, saving=saving, dir=dir, refdir=refdir)
+
+            test = bp.ansa_longitude(planet+':ansa', 'sun')
+            show_info(bp, 'Ansa longitude wrt Sun (deg)', test*DPR,
+                      printing=printing, saving=saving, dir=dir, refdir=refdir)
+
+            test = bp.ansa_longitude(planet+':ansa', 'sha')
+            show_info(bp, 'Ansa longitude wrt SHA (deg)', test*DPR,
+                      printing=printing, saving=saving, dir=dir, refdir=refdir)
+    except:
```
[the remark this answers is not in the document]
markshowalter: OK, I get it that this is surrounding unit tests, where we might not be able to enumerate all the possible exceptions. This _might_ be a place where an unqualified "except" is OK, but only if (1) you explain this in comments, and (2) if you take some action based on the exception besides ignoring it. Right now, if not printing, you will have no idea that something went wrong. Even if printing, you don't know if this set of tests still needs some development, as the message says, or if something entirely different went wrong. Maybe look into the traceback module. In the except clause, you couls catch the exception, print out a full traceback, and then continue. This way, the user will know exactly wheat went wrong, and exactly where, but it will not stop the unit tests from proceeding.

hosts/juno/junocam.py
markshowalter: The extraneous parentheses are still there. Also, BTW, when you have this many arguments in a call, I recommend you use "keyword=value" for more of them. For example, I have no idea what 'JUNO' and 'JUNO_JUNOCAM' mean in this list. As I have learned the hard way, a call like this is likely to break if anything about the call signature ever changes.

markshowalter (changes requested):
I should have been clearer that when I described my previous comments as "stylistic", I didn't mean that they were also optional. Please fix these. BTW, a global search for "\n\n\n\n" might be useful.

For Rob, I wonder if we should add to the style guide something about avoiding more than 2 or 3 function arguments without using "name=value". I am a guilty party on this. I know that Python 3 let's you specify which inputs can only be specified by name, so as we move away from Python 2 in this repo, we should adopt that more widely. Also, the style guide should probably say more about bare "except" clauses, where I am also sometimes a guilty party but have been fixing them wherever they appear now.

PR #23, opened 2022-12-13
big polymath replacement, plus two minor oops bugfixes

markshowalter:
The two changes in oops ought to be no-brainers. Regarding polymath, it might not be useful to look at the source code changes side-by-side because there are so many. However, unit tests all pass (polymath, oops, hosts) so maybe just try it at your end to make sure everything still works. Code commentary always welcome, of course!

### DIFF
--- a/oops/fov/polynomialfov.py
+++ b/oops/fov/polynomialfov.py
@@ -195,8 +195,12 @@ class PolynomialFOV(FOV):
         order = coefft.shape[0]-1
 
         (p,q) = pq.to_scalars()
-        p = p.vals[..., np.newaxis]
-        q = q.vals[..., np.newaxis]
+        if p.shape:
+            p = p.vals[..., np.newaxis]
+            q = q.vals[..., np.newaxis]
+        else:
+            p = p.vals
+            q = q.vals
 
         # Construct the powers of line and sample
         p_powers = [1.]
--- a/oops/lightsource.py
+++ b/oops/lightsource.py
@@ -53,7 +53,7 @@ class LightSource(object):
         self.source = None
         try:
             pair = Pair.as_pair(source)
-        except ValueError:
+        except (ValueError, TypeError):
             pass
         else:
             (ra, dec) = pair.values * RPD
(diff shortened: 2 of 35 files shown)


jnspitale:
All unit tests work.  There are still many relative imports.  I have created a new issue about that.

rfrenchseti:
The polymath change breaks existing oops code:

```
    ra = center_bp.right_ascension(apparent=False).vals.astype('float')
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'float' object has no attribute 'astype'
```

In this code, `center_bp` is a Backplane that covers a single point (using a Meshgrid with a single point). Previously the result of evaluating this backplane was an array with a `.vals` attribute, but now it appears to be a single float. I think this is a bad change because it means whoever uses the result of a Backplane operation needs to check to see if the result is an array or a single float if they don't know what Meshgrid was used.

rfrenchseti (changes requested):

markshowalter:
> The polymath change breaks existing oops code:
> 
> ```
>     ra = center_bp.right_ascension(apparent=False).vals.astype('float')
>          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'float' object has no attribute 'astype'
> ```
> 
> In this code, `center_bp` is a Backplane that covers a single point (using a Meshgrid with a single point). Previously the result of evaluating this backplane was an array with a `.vals` attribute, but now it appears to be a single float. I think this is a bad change because it means whoever uses the result of a Backplane operation needs to check to see if the result is an array or a single float if they don't know what Meshgrid was used.

This does not appear to be an error in oops. It is perfectly fine for the .vals property to be a scalar float--always has been. Furthermore, we know that right_ascension backplanes contain floating-point numbers, so why would one apply .astype("float") to the .vals property? Either it is an array of dtype=float already, or else it is a scalar in which case ".astype" will fail.

rfrenchseti (approved):
Single-value backplanes now return a Python float instead of a numpy.float64. Someone should fix the bad indent at some point.

PR #70, opened 2023-03-21
Deprecate backplane exercises

jnspitale:
gold_master modified to support the creation of a standard_obs.py file for a host.  This is implemented for junocam.  

I created gold masters for an adiitional junocam image, but I get an error when I try to run the tests.  See ansa.py line 316.  The error is caused by center_distance returning a full grid of values, but I think it should only have one value.


jnspitale:
Still getting this error in deprecate_backplane exercises.  We talked about it last wekk but I foget whether we resolved it...

  File "/home/spitale/pds-/pds-oops/oops/backplane/gold_master.py", line 1141, in run_tests
    test_suite(self)

  File "/home/spitale/pds-/pds-oops/oops/backplane/limb.py", line 257, in limb_test_suite
    bpt.compare(dalt.wod/bpt.duv, dalt_du,

  File "/home/spitale/pds-/pds-oops/oops/backplane/gold_master.py", line 1310, in compare
    self._compare(array, master, comparison)

  File "/home/spitale/pds-/pds-oops/oops/backplane/gold_master.py", line 1577, in _compare
    self._log_comparison(comparison)

  File "/home/spitale/pds-/pds-oops/oops/backplane/gold_master.py", line 1990, in _log_comparison
    message += [self._valstr(comparison.limit)]

  File "/home/spitale/pds-/pds-oops/oops/backplane/gold_master.py", line 2037, in _valstr
    formatted = '%#.4g' % value

  TypeError only size-1 arrays can be converted to Python scalars

jnspitale:
Highlights:
 - Added Junocam gold master tests; they work at full resolution and 16x undersampled.
 - Fixed bug where failed gold master test did not trigger unit test failure because LOGGING.errors gets reset before the error count for a given test suite is recorded.
 - With --fullpaths, changed log to have newline after every entry, not just those for which paths are logged.  Now every block of output corresponds to a single test, whether or not it is a gold-master test.

oops/backplane/gold_master.py
```diff
@@ -1048,6 +1048,7 @@ def __init__(self, obs, planets, moons, rings, overrides, args, suffix=''):
         self.gold_summary_ = None
         self.summary = {}
         self.results = {}
+        self.errors = 0
```
markshowalter, line 1051: BackplaneTest no longer has an attribute "errors", right? Delete.

oops/surface/__init__.py
```diff
@@ -18,7 +18,6 @@ class Surface(object):
     necessarily rectangular, in which two primary coordinates define locations
```
markshowalter, line 18: This is the erroneous version of surface/__init__.py with all the blank lines remove. Please replace with the previous version. This file should not have any changes.

markshowalter (changes requested):
Important, before we go much further--surface/__init__.py has been clobbered and needs to be reverted back to the version on master branch. (This is the copy in which all blank lines have been accidentally removed.) Also, I need to know the meaning of the changes to voyager/iss.py and the three cassini instruments. Are these functional changes? I don't see any gold master files for these instruments so I am wondering why the files have changed, and whether this is intentional.

jnspitale:
Just pushed surface__init__.py.  Still don't know what happened, but it appears to have affected only the doc strings.  Also removed errors attribute in gold_master.py.  

The changes to the cassini and voyager files are not functional.  Old exercise_backplanes tests were commented out and/or replaced with early versions of gold_master tests, which are now obsolete and need to be updated.   Also, relative imports were changed to absolute in cassini/iss.py

markshowalter (approved):
Looks good!

rfrenchseti (approved):

rfrenchseti:
All unit tests pass.

PR #72, opened 2023-03-28
Pds3

jnspitale:
This pull request adds a single file with some PDS3 label management methods and should not affect anything else.


rfrenchseti:
There are four other files added that start with "~". I assume these are backup files that you didn't mean to check in.

jnspitale:
removed backup files

hosts/pds3.py
```diff
@@ -0,0 +1,113 @@
+################################################################################
+# oops/inst/pds3.py
```
markshowalter, line 2: Should be hosts/pds3.py

hosts/pds3.py
markshowalter, line 1: Code looks clean and readable.

My one general comment is that I don't quite see the need for the PDS3 class per se. There are times when an instance-free class makes sense, such as inside the other host trees where we make use of inheritance a lot. However, this is more like julian.py---take a look if you are not yet familiar---where it is just a list of related functions that can all be imported at once. So the change is to remove the class, shift everything inside it left by four characters, and remove the "@staticmethod" lines. To use, "from hosts import pds3" and then refer to the functions as pds3.clean_label, etc.

hosts/pds3.py
```diff
@@ -0,0 +1,113 @@
...
+import pdsparser
+from pathlib import Path
+
+#===============================================================================
+class PDS3(object):
+    """An instance-free class for PDS3 label management methods."""
+
+    #===========================================================================
+    @staticmethod
+    def clean_label(lines):
```
markshowalter, line 17: Because the content of clean_label is going to depend on the quirks of each data set, this makes me a bit nervous. I could imagine a situation where code that cleans up one label would break another. Here's a better way: Add a new input "cleaner=None" to get_label. Inside get_label,
```
if cleaner is not None:
    lines = cleaner(lines)
```
So now "cleaner" is an optional function that you can define inside any of the instrument modules. If you pass it to get_label, that function will be used to edit the lines after they are read and before they are parsed. You would then need to define tiny cleaner functions for Cassini VIMS and UVIS, but nothing instrument-specific would show up here.

If there were a set of very common PDS3 syntax errors, then it would make a bit more sense to have a dedicated function like this one (in addition to, not instead of, the instrument-specific functionns). For example, _maybe_ it's worth a general check to make sure "N/A" is always surrounded by quotes, because I have seen that error before. ((However, if we did that test by default, we probably wouldn't make the assumptions about the adjacent parentheses that appear inside these VIMS tests.) That's really a judgment call.

markshowalter (changes requested):

hosts/pds3.py
jnspitale: My original intent was to hide all that stuff in this module, but you're right that I can't guarantee that some of these corrections would not break label from some other host.  So I did as you suggest, but I added the ability to have built-in standard cleaners that get called by default along with any custom cleaners provided by the host module.   Also, I have the cleaners operating on a single line to simplify the host code.

I have not tested the custom cleaners in order to .

hosts/pds3.py
```diff
@@ -0,0 +1,172 @@
...
+        filespec = 'cassini/VIMS/v1793917030_1.lbl'
+        label_dict = pds3.get_label(os.path.join(TESTDATA_PARENT_DIRECTORY, filespec),
+                                    cleaners=[clean_VIMS])
+        self.assertTrue(label_dict['PDS_VERSION_ID'] == 'PDS3')
+
+        # Test Cassini UVIS file
+        filespec = 'cassini/UVIS/HSP2014_197_21_29.LBL'
+        label_dict = pds3.get_label(os.path.join(TESTDATA_PARENT_DIRECTORY, filespec),
+                                                 cleaners=[clean_UVIS])
+        self.assertTrue(label_dict['PDS_VERSION_ID'] == 'PDS3')
```
markshowalter, line 167: Similar to what I have said above, the unit tests for each cleaner function belong inside that instrument module, not here.

hosts/pds3.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    Find the PDS3 label, clean it, load it, and parse into dictionary.
+
+    Inputs:
+        filespec:       File specification for the label.
+        cleaners:       List of functions that take a label line as input and
+                        return a cleaned line.  By default, these functions
+                        are appended to a list of standard cleaner functions.
+        no_standard:    If True, the standard cleaners are not used.
+    """
```
markshowalter, line 95: If the default cleaner does something you might not want, then it shouldn't do it. So I recommend you remove that option.

Why do we need a list of cleaners? A single function, which can be defined inside the instrument module, ought to suffice.

Also, because the work of a cleaner function might have to operate across multiple lines, the function should take the full list of strings and return another list of strings. It isn't necessarily possible to correct a label if the lines are only handled one at a time.

markshowalter (changes requested):

hosts/pds3.py
```diff
@@ -0,0 +1,172 @@
+################################################################################
+# hosts/pds3.py
+#
+# Utility functions for finding and loading PDS3 labels in host modules.
+#
+################################################################################
```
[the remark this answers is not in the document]
jnspitale: Extraneous lines removed

hosts/pds3.py
jnspitale: The purpose of mulitple cleaners was to be able to select which standard cleaners you wanted to use if you didn't want to use all of them.  I just pulled that stuff out and we can let the hosts do all the cleaning.  If we wind up with a lot of duplicated code, we can add functions to pds3 to do those things.

hosts/pds3.py
```diff
@@ -0,0 +1,172 @@
...
+# UNIT TESTS
+################################################################################
+import unittest
+import os.path
+
+from hosts                   import pds3
+from oops.unittester_support import TESTDATA_PARENT_DIRECTORY
+
+
+#===========================================================================
```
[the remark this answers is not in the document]
jnspitale: I have left these in for now, so that the code can be tested without this pull requiest affecting anything else.  Once we merge this file I'll move the tests to more approriate modules.

hosts/pds3.py
jnspitale: see above comment

markshowalter (approved):

PR #73, opened 2023-03-29
Voyager fixes

rfrenchseti:
My previous Voyager changes disappeared at some point so here they are again.


jnspitale (approved):
We don't currently have gold masters for voyager iss, so I can't do comprehensive testing, but I ran an emission angle backplane and it now runs and looks correct.  Also, I inspected the individual changes and tested them line-by-line and they look correct.

Issue #50, opened 2023-02-15
General cleanup of host module code


jnspitale:
This issue overlaps issues #43 and #49

jnspitale:
Tasks:
 [ ] Bring from_index() up to date
 [ ] Consistent internal structure among modules
 [ ] Document all arguments
 [ ] Consistent calling interface

jnspitale:
This issue overlaps with #159

PR #79, opened 2023-04-24
Gll ssi

jnspitale:
This should put Galileo SSI on par with Junocam: it's functional, but needs cleanup, which is addressed in issue #50.  Included are some bug fixes to gold_master and oops.  Also, note that I commented out the try-except in gold_master.execute_as_unittest().  The large try block makes it difficult to trace problems in that code section and it's not clear to me why it's needed, as it seems to just replace useful error messages with one that doesn't give much info.


jnspitale:
Removed back_plane_exercises from ideas/

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+import spicedb
+import cspyce
+import oops
+
+from oops.body import Body
+
+TOUR = (2003 - 2000) * 365 * 86400      # Rough ET dividing Saturn from Jupiter
+
+################################################################################
+# Routines for managing the loading of C and SP kernels
```
rfrenchseti, line 22: SP -> SPK?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+################################################################################
+
+# Make sure the leap seconds have been loaded
+oops.spice.load_leap_seconds()
+
+# We load CK and SPK files on a very rough month-by-month basis. This is simpler
```
rfrenchseti, line 28: This comment seems left over from Cassini, which had a tour 1997->2017.

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+# available.
+
+################################################################################
+
+class Galileo(object):
+    """An instance-free class to hold Galileo-specific parameters."""
+
+    START_TIME = '1989-10-18'
+    STOP_TIME  = '2003-09-21'
+    MONTHS = 240        # 20 years * 12 months/year
```
rfrenchseti, line 46: Why 20 years for a 14 year tour?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+    SPK_DICT    = {}
+
+    loaded_instruments = []
+
+    initialized = False
+
+    ############################################################################
+
+    @staticmethod
+    def initialize(ck='reconstructed', planets=None, asof=None,
```
rfrenchseti, line 69: Galileo doesn't have reconstructed vs. predicted kernels, or gapfill kernels. We don't have these parameters for Voyager, so they should be removed here.

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+        spicedb.open_db()
+
+        spk = spk.upper()
+        if spk == 'NONE':
+
+            # This means no SPK will ever be loaded; handling is manual
+            Galileo.initialize_kernels([], Galileo.SPK_LIST)
+            Galileo.SPK_LOADED = np.ones(Galileo.MONTHS, dtype='bool')
+        else:
+            kernels = spicedb.select_spk(-77, name='CAS-SPK-' + spk,
```
rfrenchseti, line 114: This code is dead

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+            Galileo.initialize_kernels(kernels, Galileo.SPK_LIST)
+
+        ck = ck.upper()
+        if ck == 'NONE':
+
+            # This means no CK will ever be loaded; handling is manual
+            Galileo.initialize_kernels([], Galileo.CK_LIST)
+            Galileo.CK_LOADED = np.ones(Galileo.MONTHS, dtype='bool')
+        else:
+            kernels = spicedb.select_ck(-77, name='CAS-CK-' + ck,
```
rfrenchseti, line 127: This code is dead.

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+                                             time=(Galileo.START_TIME,
+                                                   Galileo.STOP_TIME),
+                                             asof=asof)
+            Galileo.initialize_kernels(kernels, Galileo.CK_LIST)
+
+        # Load extra kernels if necessary
+        if gapfill and ck not in ('PREDICTED', 'NONE'):
```
rfrenchseti, line 134: This code is not appropriate for Galileo

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+        kdir = os.path.join(TESTDATA_PARENT_DIRECTORY, '../SPICE/Galileo/')
+        gkdir = os.path.join(TESTDATA_PARENT_DIRECTORY, '../SPICE/General/')
+
+        cspyce.furnsh(gkdir + 'LSK/naif0012.tls')
+        cspyce.furnsh(kdir + 'SCLK/mk00062a.tsc')
+        cspyce.furnsh(kdir + 'IK/gll36001.ti')
+        cspyce.furnsh(kdir + 'SPK/de421.bsp')
+        cspyce.furnsh(kdir + 'SPK/de432s.bsp')
+
+        [cspyce.furnsh(ckfile) for ckfile in glob.glob(kdir + 'CK/*.bc')]
```
rfrenchseti, line 219: I'm not fond of having a list comprehension with side effects. I think this would be a lot more readable as a for loop.

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+        [cspyce.furnsh(spkfile) for spkfile in glob.glob(kdir + 'SPK/*.bsp')]
+
+
```
rfrenchseti, line 222: Remove extra blank line

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+        return
+
+## TODO:
```
rfrenchseti, line 225: Is there still something TODO here?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+                    kernel_dict[filespec] = kernel
+                loaded[m] = True
+
+    ############################################################################
+    # Initialize the kernel lists
+    ############################################################################
+
+    @staticmethod
+    def initialize_kernels(kernels, lists):
+        """After initialization, lists[m] is a the KernelInfo objects needed
```
rfrenchseti, line 249: is a the? objects?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+
+            m1 = int((t0 - Galileo.TDB0) // Galileo.DTDB)
+            m2 = int((t1 - Galileo.TDB0) // Galileo.DTDB) + 1
+
+            m1 = max(m1, 0)     # ignore time limits outside mission duration
+            m2 = min(m2, Galileo.MONTHS - 1)
+
+            # Add this kernel to each month's list
+            for m in range(m1, m2+1):
+                lists[m] += [kernel]
```
rfrenchseti, line 269: Maybe just use append?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+
+    ############################################################################
+    # Routines for managing the loading other kernels
```
rfrenchseti, line 272: loading of other?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+    ############################################################################
+
+    @staticmethod
+    def load_instruments(instruments=[], asof=None):
+        """Load the SPICE kernels and defines the basic paths and frames for
```
rfrenchseti, line 277: defines -> define

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+        the Galileo mission.
+
+        It is generally only to be called once.
+
+        Input:
+            instruments an optional list of instrument names for which to load
+                        frames kernels. The frames for ISS, VIMS, UVIS, and CIRS
```
rfrenchseti, line 284: Old comment for Cassini

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+
+            asof        if this specifies a date or date-time in ISO format,
+                        then only kernels that existed before the specified date
+                        are used. Otherwise, the most recent versions are always
+                        loaded.
+        """
+
+        # Load the default instruments on the first pass
+        if Galileo.loaded_instruments == []:
+           instruments += ['SSI']
```
rfrenchseti, line 295: append?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+        # Furnish instruments and frames; Note GLL has no frame kernel
+        spicedb.open_db()
+        _ = spicedb.furnish_inst(-77, inst=instruments, asof=asof)
+        spicedb.close_db()
+
+    ############################################################################
+    # Routines for managing text kernel information
+    ############################################################################
+
+### TODO: finish these routines...
```
rfrenchseti, line 315: Is there still TODO here?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+
+    @staticmethod
+    def spice_instrument_kernel(inst, asof=None):
+        """A dictionary containing the Instrument Kernel information.
```
rfrenchseti, line 319: Return a dictionary?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+
+        Also furnishes it for use by the SPICE tools.
+
+        Input:
+            inst        one of "ISS", "UVIS", "VIMS", "CIRS", etc.
```
rfrenchseti, line 324: Old comments

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+        kernel_info = spicedb.select_inst(-77, types='IK', inst=inst, asof=asof)
+        spicedb.furnish_kernels(kernel_info, fast=True)
+        spicedb.close_db()
+
+        return (spicedb.as_dict(kernel_info), spicedb.as_names(kernel_info)[0])
+
+    #===========================================================================
+    @staticmethod
+    def spice_frames_kernel(asof=None):
+        """A dictionary containing the Galileo Frames Kernel information.
```
rfrenchseti, line 348: Return a dictionary?

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+    @staticmethod
+    def used_kernels(time, inst, return_all_planets=False):
+        """The list of kernels associated with a Galileo observation at a
+        selected range of times.
+        """
+        if return_all_planets:
+            bodies = [1, 199, 2, 299, 3, 399, 4, 499, 5, 599, 6, 699,
+                      7, 799, 8, 899]
+            if time[0] >= TOUR:
+                bodies += Body.JUPITER_MOONS_LOADED
```
rfrenchseti, line 383: This code doesn't make any sense

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+import pdsparser
+import oops
+
+from hosts         import pds3
+from hosts.galileo import Galileo
+
+################################################################################
+# Standard class methods
+################################################################################
+def from_file(filespec, fast_distortion=True,
```
rfrenchseti, line 18: fast_distortion is left over from Cassini ISS

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+                               basename = os.path.basename(filespec))
+
+    result.insert_subfield('spice_kernels',
+                           Galileo.used_kernels(result.time, 'iss',
+                                                return_all_planets))
+
+    return result
+
+#===============================================================================
+def initialize(ck='reconstructed', planets=None, asof=None,
```
rfrenchseti, line 78: Same comments about ck, spk, gapfill as arguments and in comments

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+        irregulars  True to include the irregular satellites;
+                    False otherwise.
+    """
+    SSI.initialize(ck=ck, planets=planets, asof=asof,
+                   spk=spk, gapfill=gapfill,
+                   mst_pck=mst_pck, irregulars=irregulars)
+
+
+#===============================================================================
+class Metadata(object):
```
rfrenchseti, line 106: Maybe call this GalileoSSIMetadata or something? Metadata seems awfully generic.

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+        Input:
+            label           The label dictionary.
+
+        Attributes:
+            nlines          A Numpy array containing the data in axis order
+                            (line, sample).
+            nsamples        The time sampling array in (line, sample) axis
+                            order, or None if no time backplane is found in
+                            the file.
+            nframelets
```
rfrenchseti, line 121: Comment?

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+        """Use the label to assemble the image metadata.
+
+        Input:
+            label           The label dictionary.
+
+        Attributes:
+            nlines          A Numpy array containing the data in axis order
+                            (line, sample).
+            nsamples        The time sampling array in (line, sample) axis
+                            order, or None if no time backplane is found in
```
rfrenchseti, line 119: Does Galileo have a time backplane?

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+            self.window = None
+
+    #===========================================================================
+    def trim(self, data, full_fov=False):
+        """Trim image to label window
+
+        Input:
+            full_fov        If True, the image is not trimmed.
+
+        Attributes:
```
rfrenchseti, line 160: I don't think the Attributes part of this comment is relevant

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+
+        if full_fov:
+            return None
+
+        window = self.window
+
+        if window is None:
+            return data
+
+        return data[window[0]:window[2], window[1]:window[3]]
```
rfrenchseti, line 178: I'm not sure of the Galileo convention, but are the window dimensions non-inclusive of the upper limit?

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+
+    #===========================================================================
+    def fov(self, full_fov=False):
+        """Use the label to assemble the image metadata.
```
rfrenchseti, line 182: Entire comment not relevant

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+        cxy_var = 'INS-77036_FOV_CENTER'
+
+        cf = cspyce.gdpool(cf_var, 0)[0]
+        fo = cspyce.gdpool(fo_var, 0)[0]
+        px = cspyce.gdpool(px_var, 0)[0]
+        cxy = cspyce.gdpool(cxy_var, 0)
+
+        # Construct FOV
+        scale = px/fo
+        distortion_coeff = [1,0,cf]
```
rfrenchseti, line 212: Add spaces after commas

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+
+        # Direct summation modes
+        if self.mode=='HIS' or self.mode=='AI8':
```
rfrenchseti, line 215: Add spaces around ==. Could also just say `if self.mode in ('HIS', 'AI8'):`

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+class SSI(object):
+    """An instance-free class to hold Galileo SSI instrument parameters."""
+
+    instrument_kernel = None
+    fov = {}
+    initialized = False
+
+    #===========================================================================
+    @staticmethod
+    def initialize(ck='reconstructed', planets=None, asof=None,
```
rfrenchseti, line 248: Same comments

rfrenchseti (changes requested):
I've commented on the code but not yet run the tests. I'll wait until these comments are addressed to do a functionality test.

oops/backplane/gold_master.py
```diff
@@ -578,46 +578,46 @@ def execute_as_unittest(testcase, obsname='default'):
...
+    args.task = 'compare'
+    args.level = 'error'
+    args.verbose = True
+    args.du = 0.
+    args.dv = 0.
+    args.derivs = True
+
+    # Clean up, also filling in observation, module, planet(s), moon(s),
+    # ring(s)
+    args = _clean_up_args(args)
```
markshowalter, line 617: As we discussed, restore the previous try/except. Then, before testcase.assertTrue(False), you can print the exception and traceback by adding `traceback.print_exc()`

markshowalter (changes requested):
I will start my comments on the galileo module shortly. These are comments about the oops changes, just to get you started. A few we already discussed.

oops/path/spicepath.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, spice_id, spice_origin="SSB", spice_frame="J2000",
         self.spice_frame = spice_frame
         self.shortcut = shortcut
 
-        # Interpret the SPICE IDs
+       # Interpret the SPICE IDs
```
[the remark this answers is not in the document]
jnspitale: Corrected

hosts/galileo/__init__.py
jnspitale: K stands for kernel, so CK / SPK kernel is redundant

hosts/galileo/__init__.py
jnspitale: Yeah, the dates correspond to Cassini.  The approach is the same, so I'll just change the dates.  I don't think it's worth worrying about beyond that since we're going to re-do the spice handling anyway.

hosts/galileo/__init__.py
jnspitale: fixed

hosts/galileo/__init__.py
jnspitale: True, but also probably not worth worrying about right now.

hosts/galileo/__init__.py
jnspitale: Ok, I just went ahead and pulled all that stuff out.

hosts/galileo/__init__.py
jnspitale: removed

hosts/galileo/__init__.py
jnspitale: fixed

hosts/galileo/__init__.py
jnspitale: nope, deleted

hosts/galileo/__init__.py
jnspitale: fixed

hosts/galileo/__init__.py
jnspitale: Is there a benefit?  This is how it's done in the cassini module.

hosts/galileo/__init__.py
jnspitale: Seems right

hosts/galileo/__init__.py
jnspitale: fixed

hosts/galileo/__init__.py
jnspitale: Removed entire method

hosts/galileo/__init__.py
jnspitale: same as above comment.  Cassini does it this way.  Does it matter?

hosts/galileo/__init__.py
jnspitale: nope, removed th methods

hosts/galileo/__init__.py
jnspitale: Agreed, this is the result of a search and replace.  This was one of the TODO methods at the end of the file that I have deleted in response to an earlier comment.  As far as I can tell, although the output of this method is added to the observation, it is never used for anything (even in Cassini ISS), and it's not clear what it's purpose is, because it doesn't actually reflect which kernels are used.

Indeed, I don't currently use this in Juno, so deleteing it here is in the spirit of  'on par with Junocam'.

hosts/galileo/ssi/__init__.py
jnspitale: Couldn't the same be said about from_file()?  If anyone ever wanted access this from outside the Galileo SSI module, they would have to qualify it.  But maybe it should be _Metadata, since it is not meant to be accessed externally.

hosts/galileo/ssi/__init__.py
jnspitale: removed

hosts/galileo/ssi/__init__.py
jnspitale: This comment is not even relevant; removed

hosts/galileo/ssi/__init__.py
jnspitale: Agreed, removed.

hosts/galileo/__init__.py
jnspitale: I agree this is a useful thing to have, but the list returned by this function has no connection to which kernels were actually used to create the observation.  What is really needed is for whatever function that loads the kernel to record their names at that time.  This could be done now, but will likely be part of the larger kernel management project.

hosts/galileo/ssi/__init__.py
jnspitale: Yes, many of the comments we are discussing need to be addressed as part of the larger issue of cleaning up all of the host modules.  So I'm going to leave this one alone unless you think there's a reason to deal with it now for just this module.

hosts/galileo/__init__.py
rfrenchseti: I don't know why you say that. It works great for Cassini ISS. Why wouldn't it work for Galileo? Yes, it includes all of the moon kernels, but the important part is which pointing kernels were used.

This code returns the kernels needed for the time range of the observation:

```
        return spicedb.used_basenames(time=time, inst=inst, sc=-77,
                                      bodies=bodies)
```

hosts/galileo/__init__.py
jnspitale: Just realized the isuue is append() appends a list as a single item, whereas + appends each item as a separate list item, which is the desired behavior.  In other words:

[1,2,3].append([4,5]) is [1,2,3, [4,5]], 

while

[1,2,3] + [4,5] is [1,2,3,4,5]

rfrenchseti: Right, but `kernel` is a single object. `[kernel]` makes a list of length one with kernel as the only element. So these are equivalent:
```
lists[m] += [kernel]
lists[m].append(kernel)
```

jnspitale: True, makes sense if appending a single object.

hosts/galileo/__init__.py
jnspitale: Restored and updated used_kernels() for galileo.

hosts/galileo/ssi/__init__.py
jnspitale: It is, but this code is still incorrect.  First, it's (line,sample), not (u,v); second, the second pair of numbers is the size of the region, not the corner.  All of the test images used so far start at (1,1), so the second pair of numbers is both the corner coordniate and the size of the region.  I just looked at some others and confirmed this.

jnspitale: I added a gold-master image with a cutout window far from the origin and it looks like this is correct now.

jnspitale:
All comment addressed, ready for additional review.

hosts/galileo/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+        spicedb.close_db()
+
+    ############################################################################
+    # Routines for managing text kernel information
+    ############################################################################
+
+    #===========================================================================
+    @staticmethod
+    def used_kernels(time, inst, return_all_planets=False):
+        """The list of kernels associated with a Cassini observation at a
```
rfrenchseti, line 284: Cassini -> Galileo

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+# Standard class methods
+################################################################################
+def from_file(filespec,
+              return_all_planets=False, full_fov=False, **parameters):
+    """A general, static method to return a Snapshot object based on a given
+    Galileo SSI image file.  By default, only the valid image region is
+    returned.
+
+    Inputs:
+        full_fov:           If True, the full image is returned with a mask
```
rfrenchseti, line 25: filespec is also an input (yes I know it's missing from Cassini ISS). Order of other two comments should be swapped.

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+
+        return_all_planets  Include kernels for all planets not just
+                            Jupiter or Saturn.
+    """
+
+    SSI.initialize()    # Define everything the first time through; use defaults
+                        # unless initialize() is called explicitly.
+
+    # Load the PDS label
+    label = pds3.get_label(filespec)
```
rfrenchseti, line 36: Is it possible to load an image without requiring the label, perhaps with limited functionality? We don't generally require the label for other instruments.

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+            self.window_uv_shape = np.flip(self.window_shape)
+        else:
+            self.window = None
+
+    #===========================================================================
+    def trim(self, data, full_fov=False):
+        """Trim image to label window.
+
+        Input:
+            full_fov        If True, the image is not trimmed.
```
rfrenchseti, line 139: data is also an input

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+
+        Output:
+            Data array trimmed to the data window.
+        """
+        if full_fov:
+            return data
+
+        window = self.window
+        if window is None:
```
rfrenchseti, line 148: window isn't used again so just say `if self.window is None`

hosts/galileo/ssi/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+        Galileo.reset()
+
+################################################################################
+# UNIT TESTS
+################################################################################
+import unittest
+import os.path
+import oops.backplane.gold_master as gm
+
+from oops.unittester_support    import TESTDATA_PARENT_DIRECTORY
```
rfrenchseti, line 269: Remove extra spaces

rfrenchseti (changes requested):
Looks good. Only a few minor comments. However, the gold-master tests for Galileo are failing for me.

hosts/galileo/ssi/__init__.py
jnspitale: corrected in cassini, juno, and voyager as well

hosts/galileo/ssi/__init__.py
jnspitale: We should be able to get everything from the vicar label.  I suggest moving forward with the current code right now and I will make an issue to re-do this using the VICAR label.

jnspitale:
I added overrides to allow the gold masters to work for both of you, so go ahead and try again.  I notice there were no errors from the last test, which is new, so you may get errors from that one.

rfrenchseti:
Unit tests now pass for me.

rfrenchseti (approved):

PR #82, opened 2023-05-08
JWST and navigation

markshowalter:
I am opening this pull request now because it relates to Rob's recent question about image navigation. There's no rush on this so it can wait till after GLL SSI is merged. In addition to a lot of JWST/NIRCam functionality, this provides a lot of new support for image navigation. As an example, I added two new optional parameters to `voyager.iss.from_file()` called `navigation` and `offset`. The former lets you specify the corrections to the C kernel as rotation angles and the latter as pixel offsets. In my own data analysis, I have been saving dictionaries of these corrections and then applying them immediately when I call `jwst.nircam.from_file()`. Alternatively, calling `from_file()` with `naviation=True` returns the Observation object with an embedded Navigation frame, which makes it easier to solve for a pointing correction iteratively/interactively.

The support for this capability ripples down through various other modules:
- There's a minor revision to the `frame/navigation.py` constructor.
- New methods `fov.offset_angles_from_duv()`, `fov.offset_duv_from_angles()` convert between pixel offsets and Navigation rotation angles.
- There are four new methods to support "parallel" observations in `observation/__init_.py`: `parallel_los()`, `parallel_uv()`, `parallel_offset_angles()`, `parallel_offset_duv()`.
- New method `Vector3.offset_angles()` and unit test additions for `Vector3.spin()`.

In addition:
- IMPORTANT bugfix in `hosts/solar/__init__.py` where previously, certain function inputs were ignored.
- `Observation.uv_from_path()` now works for many TimedImage objects (such as NIRCam).
- Bugfix and revised unit tests for `Scalar.solve_quadratic()`.
- LOTS of revisions to `hosts/jwst`.


hosts/voyager/iss.py
```diff
@@ -142,6 +142,25 @@ def from_file(filespec, astrometry=False, action='error', parameters={}):
                           label_dict['PRODUCT_ID'], RuntimeWarning)
             image_frame = spacecraft + '_ISS_' + camera
 
+    # Wrap the frame in a Navigation if necessary
```
jnspitale, line 145: Is this idea intended to be applied to all hosts eventually?  If so, I'll create an issue.

oops/backplane/lighting.py
```diff
@@ -409,7 +409,8 @@ def lighting_test_suite(bpt):
 
         # Select the 95th percentile as a large but representative value
         # This is needed because km_per_fov_radian diverges near limb
-        values = deg_per_fov_radian.vals[deg_per_fov_radian.antimask]
+        values = deg_per_fov_radian[deg_per_fov_radian.antimask].vals
```
jnspitale, line 412: This is still going to cause a problem because the result ot np.sort() has the same shape as the input, so it still needs to be flattened at some point before indexing it with the cutoff index.  This doesn't trip a unit test error in this pull request because the test that triggers it occurs inthe galileo ssi pull request.

oops/observation/__init__.py
```diff
@@ -955,6 +951,122 @@ def inventory(self, bodies, tfrac=0.5, time=None, expand=0.,
         raise NotImplementedError(type(self).__name__ + '.inventory '
                                   'is not implemented')
 
+    ############################################################################
+    # Support for parallel observations
```
jnspitale, line 955: I find the word "parallel" confusing in this context because it has a geometric connotation implying that the lines of sight are the same.   I would pfrefer "coupled" instead.

"Simultaneous" or "concurrent" are also ok, but they really only imply the same time, not the same place ( and "simultaneous" doen't even imply the same time in relativity).

jnspitale (changes requested):
Oops and host unit tests pass, but they don't test the changed code.

hosts/voyager/iss.py
markshowalter: Yes, this is the idea. The other way to do it would be to have an Observation method called "navigate()", perhaps, that does this after the Observation was already created. That may be better because it would not have to be implemented for every host subclass. Regardless, we need an issue that says "figure out how to implement navigations of Observation objects".

oops/backplane/lighting.py
markshowalter: I'm changing this to np.sort(values, axis=None), because axis=None tells it to sort across all axes, yielding a 1-D array no matter the shape. Note that values.flatten() makes a copy first, which is unnecessary in this case. np.sort(values.ravel()) would be similar to axis=None, because ravel only makes a copy if it needs to.

oops/observation/__init__.py
markshowalter: The term "parallel" is used throughout HST and JWST documentation, so I guess that is where I picked it up. I don't think the other terms you propose really capture the correct meaning. The requirements are same time, same observing platform, and using instruments that cannot be pointed independently of one another. Cassini ISS BOTSIMs and ORS riders would qualify as parallels. Let's discuss.

jnspitale: Sounds good.  It's difficult to find the exact right word.

jnspitale (approved):

rfrenchseti (changes requested):
All tests pass.

Just a couple of very minor nits.

oops/observation/__init__.py
markshowalter: I think you're both being too literal. For all practical purposes, they _are_ parallel lines of sight--multiple instruments looking at the same target at the same time from the same host, using instruments that are designed to be used in (dare I say it) parallel.

PR #94, opened 2023-09-21
Should fix the metadata bug

markshowalter:
This should fix the bug where certain surface "hints" were being misinterpreted. Most of the changes are in docstrings to be much clearer about what's going on. I did identify a few other very minor bugs and fixed them as well. All unit tests pass. BTW, the lines that said `if hints in (None, True)` never worked correctly, because 1 == True and also Scalar(1) == True. The new test is ```isinstance(hints, (type(None), bool, np.bool_))``` which does the right thing, although it also means hints=False does the same thing as hints=True, which is probably OK. Also BTW, if you were unaware, np.bool_ is not a subclass of bool, which is annoying, so I include that because somebody might conceivably pass a numpy boolean array element as a hint, and we need that to work the same way as a bool.

### DIFF
--- a/oops/surface/__init__.py
+++ b/oops/surface/__init__.py
@@ -112,9 +112,9 @@ class Surface(object):
             hints       optional data used to expedite this calculation. The
                         specific meaning depends on the Surface subclass.
 
-        Return:         a tuple containing two to four values.
-            coords      two or three coordinate values, depending on the value
-                        of axes.
+        Return:
+            coords      a tuple of two or three coordinate values, depending on
+                        the input value of axes.
         """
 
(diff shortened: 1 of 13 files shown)


jnspitale (changes requested):
Unit tests pass, fixes the bug we discussed.  My only comment is that the keyword 'hints' does not do a good job of describing its function, especially as there is often a 'guess' as well.  'quick' and 'fast' come to mind as alternatives.  'quick' seems to be consistently used wrt quickframes as a boolean input, but 'fast' is used in various contexts, and it seems like using it here would be consistent with the variety of ways it is used in other contexts, where it directs the code to use a different algorithm in one way or another.

PR #124, opened 2024-03-04
Remove Cassini ISS offset

### LINKED ISSUE #96: Investigate offsets in oops hosts
Various hosts (Cassini ISS, New Horizons LORRI) have hard-coded offsets. Decide whether we want to include offsets in hosts, and if so what they should be.

rfrenchseti:
There is a NAC/WAC offset hard-coded in the Cassini ISS hosts module that I derived from previous navigation work, but I do not have confidence in this number or (based on information from Joe) the possibility that it might vary over time. Thus I'm turning it off by default until we can do further analysis. I think having the offset on by default will do more potential damage. I also removed the same copied code from JUNOCAM, even though that code is apparently never called. All tests still pass with current gold backplanes.

Fixes #96

### DIFF
--- a/oops/hosts/cassini/iss.py
+++ b/oops/hosts/cassini/iss.py
@@ -144,7 +144,7 @@ def from_index(filespec, **parameters):
     return snapshots
 
 #===============================================================================
-def initialize(ck='reconstructed', planets=None, offset_wac=True, asof=None,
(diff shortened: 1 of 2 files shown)


jnspitale (approved):
This sounds good to me.  The Junocam argument was left over from copying and modifying the Cassini module.

PR #125, opened 2024-04-08
Fix VGISS to work with new rms-vicar library, LORRI to work with oops changes

rfrenchseti:
Fix bug in `hosts.voyager.iss` caused by change in `rms-vicar` module. Fix bug in `hosts.newhorizons.lorri` caused by changes in oops directory layout.

Also fix problem running coverage report on Windows.


markshowalter (approved):
All OK.

PR #126, opened 2024-04-16
Fixed for UVIS and VIMS

markshowalter:
These are fixes to UVIS and VIMS, sufficient to change the color code in Table 1, row 2 of Joe's proposal to green. There are still no unit tests but the obvious bugs were fixed and the modules now work fine according to my informal testing. They are certainly no worse than what was in there before today, so I hope they can be merged with minimal fuss.

One change that you may appreciate is that I added a new function fast_dict() in hosts/pds3. It does a quick parsing of a PDS3 label and returns a dictionary that is remarkably similar to what you expect from pdsparser.PdsLabel.from_file().as_dict(), but is 100 times faster. That's not a typo. It's a brute-force parser without using pyparsing, but not really all that complicated. It does no validation at all, so for that you still need pdsparser. There will probably be some edge cases where it can't read a label, but it appears to be quite robust in my testing so far. It is now the reader used for PDS3 labels in the VIMS and UVIS modules and it can handle the VIMS internal ISIS labels too.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,3 +4,7 @@ coverage:
       default:
         # basic
         target: 0
+    patch:
+      default:
+        # basic
+        target: 0
--- a/oops/hosts/cassini/uvis.py
+++ b/oops/hosts/cassini/uvis.py
@@ -1,22 +1,24 @@
-################################################################################
+##########################################################################################
 # oops/hosts/cassini/uvis.py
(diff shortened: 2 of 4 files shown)


jnspitale:
Thanks, that sounds great  I'll update the table.

rfrenchseti (approved):

PR #129, opened 2024-04-19
Major reorg including separating tests into separate directory

markshowalter:
1. These files have been identifed as deprecated for some time. I have moved them to `ideas/deprecated`.
```
    oops/calibration/extendedsource.py
    oops/calibration/pointsource.py
    oops/fov/polynomialfov.py
    oops/fov/radialfov.py
    oops/hosts/cassini/iss_test_suite.py
    oops/hosts/cassini/uvis_test_suite.py
    oops/hosts/cassini/vims_test_suite.py
    oops/hosts/hst/wfc3_pointing_test.py
    oops/hosts/hst/wfc3_test_suite.py
    oops/hosts/hst/wfc3/wfc3_pointing_test.py to deprecated/wfc3_pointing_test copy.py
    oops/hosts/hst/wfc3/wfc3_test_suite.py to deprecated/wfc3_test_suite copy.py
    oops/observation/NOTES-ABOUT-OBSERVATION-CLASSES.txt
    oops/observation/pushbroom.py
    oops/observation/pushframe.py
    oops/observation/rasterscan.py
    oops/observation/rasterslit.py
    oops/observation/slit.py
```

2. Of the remaining fov subclasses, I renamed `PolyFOV` to `PolynomialFOV`. The latter is a more specific name (what does "poly" mean?). Where used in a host, it has been updated.

3. Based on discussions with Rob, I have cleaned up some of our import procedures. The primary goals were to keep things simple and make sure that users/programmers do not have to know the details of our file directory structure and naming conventions. To that end, as before, after typing:
```
    import oops
```
all of the following class references are valid:
```
    oops.Event              # for a class without subclasses
    oops.cadence.Cadence    # for an abstract class
    oops.cadence.Metronome  # for a subclass of an abstract class
```
I have now added:
```
    oops.Cadence            # alternative, for an abstract class
```
In other words, all base-level classes can be referenced by "`oops.Name`", whether abstract or not.

4. Previously, after one typed
```
    import oops
```
it would cease to be possible to import a class using its file path. This was because `oops/__init__.py` contained, for example, the line
```
    import oops.cadence.all as cadence
```
This had the side-effect of re-defining "`oops.cadence`" so that it no longer referred to a directory; it would refer to the file `oops/cadence/all.py` instead. From this point on,
```
    from oops.cadence.metronome import Metronome
```
would no longer work! That issue has now been repaired; it is now always possible to import a class via its directory path. This is accomplished by re-purposing all the `oops/name/__init__.py` files to import the subclasses, a task previously done by `oops/name/all.py`.

5. To accomplish the above, I had to relocate the abstract class definitions out of the `__init__.py` files. They are now in files `oops/name/name_.py`. So whereas previously, you could say
```
    from oops.cadence import Cadence
```
Now, you can't. Instead, you need to say
```
    from oops.cadence.cadence_ import Cadence
```
I included the trailing underscore on each of these files to make sure it is distinguished from the directory name.

6. By the way, I did not change the `oops/backplane` directory or import approach, because `Backplane` is not an abstract class; instead, it is just a class whose methods are defined in multiple files.

7. As a general rule, the only place where you really need to import classes using the file path approach is inside the core oops code, excluding `hosts/`. The reason is that this is the key place where we need to limit the import of modules so as to avoid circles. Users and host developers should never need to be specific about the locations of files; `import oops` should suffice. 

8. Unfortunately, the above change created a bunch of circular imports. I fixed this by defining a much cleaner ranking, where each of these classes can reference those later in the list but not earlier:
```
    Body, Surface, Path, Gravity, Event, Frame, Transform
```
This is documented in `oops/__init__.py`, which includes the "hooks" needed so that necessary class references all still work.

9. **BIG CHANGE**. For consistency with other RMS modules, I moved all unit tests except those in `oops/hosts/`, into a new directory tree `tests/`. The `unittest.py` files have also been moved. GitHub now runs these tests in their new location, via a one-line change to `scripts/automated_tests/oops_main_test.sh`. Part of the reason for this layout is that we've decided "pip install" shouldn't install our unit tests.

I did not move the hosts unit tests.

10. **BIG CHANGE**. I have moved the gold_master app out of `oops/backplane` and into a new home in `oops/gold_master`. It did not seem to me that it should remain "buried" inside `oops/backplane`. It is also something that we do want to come with a "pip install", so it didn't belong inside `tests/`.

11. To support the above, I had to move the gold-master-specific environment variable checks out of `oops/unittester_support.py` and into `oops/gold_master/test_support.py`. At some point, we will probably want to eliminate `unittester_support.py` from `oops/` but for now I left it there.

12. TODO. I discovered that, owing to a typo, the unit tests for `PolarLimb` were not executing. (In fact, they were executing on the `Limb` class instead.) When I made the obvious change, the tests failed. For now, `PolarLimb` is un-tested and may have bugs. The file `tests/surface/test_polarlimb.py` has been renamed to `tests/surface/xtest_polarlimb-NOT-WORKING.py` till this can be sorted out.

13. In a minor bug, I found that the backplanes `where_below`, `where_above`, and `where_between` did not support `Scalar` values. This is now fixed.


jnspitale (approved):
All unit tests pass

rfrenchseti (approved):

PR #131, opened 2024-05-28
Host unit tests moved

markshowalter:
This completes the changeover of all `oops/hosts` unit tests to `tests/hosts`. I needed it so that I could more clearly identify the code coverage of various modules once you exclude the test code itself. Summary of changes...

* In `oops/hosts`, I removed unit tests from every module. This includes `gold_master.py` and `standard_obs.py `wherever they appeared.

* Previously, `tests/hosts` was just a copy of `oops/hosts`, because that's where I had left things when I moved all the other unit tests. Now, each file just contains the unit tests. This is also where you find all files `gold_master.py,` `standard_obs.py`, and `unittester.py`. Conceptually, things are set up so that `tests/hosts/mission/instrument/unittester.py` tests the given instrument, `tests/hosts/mission/unittester.py` tests all the instruments on the mission, and `tests/hosts/unittester.py` tests all hosts.

* For my own convenience, I added `tests/unittester_with_hosts.py`, which runs all the OOPS unit tests including hosts.

* Where there were random files or unused bits of code in `oops/hosts` that had once been used for testing but are now unused, I moved them to `tests/hosts` but they are still unused.

* In `scripts/automated_tests/oops_main_test.sh`, I changed the location for the hosts tests from `oops/hosts/unittester.py` to `tests/hosts/unittester.py`.

* In `tests/*.py`, I made sure each `setUp()` function loaded the kernels that are needed for that test. The philosophy is that each test is responsible for its own kernels to be present (and must-be-excluded kernels to be absent). If a test does not do that, outcomes could be unpredictable. For the most part, `tearDown()` functions are not implemented (or I left them in there with "pass"), but I use them to undo anything especially pathological I did in the `setUp()`. 

* I added `_undefine_solar_system()` to `Body`, which wipes the internal cache; it's needed for debugging.

* Similarly, I added `_unload_all()` to `spicedb.py`, which unloads all kernels. Again, needed for debugging.

* Bugfix: the reference to `AliasFrame` in `body.py` didn't work quite as expected. I changed it to `Frame.as_frame()`.

* Bugfix:` hosts/galileo/__init__.py` was not loading the Galileo Frames kernel, which means I don't quite know how it worked before--apparently a leftover kernel load from somewhere else was hiding this bug.

* There's a small discrepancy between some gold-master calculations on Windows vs. Mac/Linux. We have seen it before and it remains unexplained. Nevertheless, I needed to update the error threshold on some Galileo Europa tests to get this to pass for Windows.  Several angles (center phase, orbital longitude) had discrepancies of 0.0012-0.0014 degrees, whereas the default error threshold was 0.001.


tests/surface/test_spheroid.py
```diff
@@ -262,8 +262,8 @@ def runTest(self):
             ref = t.d_dpos.vals[...,i]
             self.assertTrue(abs(dt_dpos/ref - 1).max() < 1.e-5)
 
-        Path.reset_registry()
-        Frame.reset_registry()
+#         Path.reset_registry()
+#         Frame.reset_registry()
```
rfrenchseti, line 266: In this and other similar files, why are these commented out instead of deleted?

rfrenchseti (approved):

PR #132, opened 2024-06-11
Cleanup of error logging in gold_master.py and Galileo SSI

markshowalter:
This pull request contains fixes to how errors are logged. It was part of the previous pull request but that one got merged before I had pushed these changes.

In the LOGGING methods in `oops/config.py`:
* Add counter that records how many lines a logger has printed.
* In `print()`, literal messages no longer get counted as warnings or errors, even if the level is WARN or ERROR. This is consistent with the point of literal messages, which is to add supplementary info to the log. Supplementary info about a warning or error should not increment the count.

In `oops/gold_master/__init__.py`:
* Add sensible default values to `define_standard_obs()`.
* Force the planets, moons, and rings inputs to be tuples, never lists.
* Bugfix: "info" was not a recognized input for --level.
* Force a log message to be printed identifying the observation in which an error is logged. Previously, if the logging level was set to something higher than INFO and the test was being run on multiple observations, an error would be logged without identifying which of multiple observations triggered the error.
* There is now a blank line in the log between multiple observations tested when the level is INFO.
* When multiple observations were tested, the log ends with INFO messages indicating the total warnings, total errors, and total elapsed time.
* When logging internals, I now ensure that `LOGGING.pop()` is always called after `LOGGING.push()` using try/finally.

For Galileo SSI:
* I removed the "extraneous byte" warning because it seems to be a common issue in the data files and we really don't care.
* I simplified the calls to `define_standard_obs()`, taking advantage of the new input defaults.

### DIFF
--- a/oops/config.py
+++ b/oops/config.py
@@ -139,6 +139,7 @@ class LOGGING(object):
     log_formatting = True           # True to use DEFAULT_LOG_FORMAT.
     warnings = 0                    # Warning count.
     errors = 0                      # Error count.
+    lines = 0                       # Number of lines logged.
     python_warnings = False         # Use Python warnings on LOGGING.warn()
 
     LEVELS = {                      # Static dictionary of logging levels
@@ -168,6 +169,7 @@ class LOGGING(object):
 
         LOGGING.warnings = 0
         LOGGING.errors = 0
+        LOGGING.lines = 0
 
     @staticmethod
     def all(flag, category='', reset=False):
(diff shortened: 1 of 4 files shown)


jnspitale (approved):
Loosk good, all unit test pass.  It looks like I miust have had some extraneous overrides in the galileo ssi standard_obs file.

markshowalter:
> Loosk good, all unit test pass. It looks like I miust have had some extraneous overrides in the galileo ssi standard_obs file.

In the third test, there were a bunch of Europa overrides, but they were applied to an Earth observation so they never got used. The Europa overrides were only needed for the fourth test.

rfrenchseti (approved):

PR #138, opened 2024-10-09
Changes to support rms-filecache

rfrenchseti:
The primary goal of this PR is to support the `rms-filecache` package, which abstracts reading and writing of files that could be on the local filesystem or in the cloud. This means that people using `spicedb` or `oops` no longer need to have a copy of the OOPS-Resources DropBox on their local disk; files are downloaded into a local cache from a remote cloud source on an as-needed basis. Cached files are kept across runs so that they don't need to be downloaded again. For examples of how to use `rms-filecache` and full documentation see: https://rms-filecache.readthedocs.io/en/latest/

You should see no changes with your existing setup. Current environment variables are supported in exactly the same way and local file accesses, although they go through `filecache` methods, still access the local filesystem unimpeded. However, if you want to use the cloud, you can do something like:

```sh
export OOPS_RESOURCES=gs://rms-node-oops-resources
```

and now all files will be downloaded from the Google Cloud `rms-node-oops-resources` bucket, which contains all of the SPICE kernels, test data, and golden masters. Note that you need to have a Google Cloud account and be authenticated locally for this to work, since the bucket is not currently publicly accessible.

- Changes to support filecache:
  - Many parts of the code now use `Path` objects instead of strings.
  - Changed `spicedb` to use `filecache`. Kernels are downloaded in parallel when possible. All tests have been updated and pass. This also fixes #137 .
  - Changed all oops unit tests to retrieve test files from the `test_data` directory and kernels from the `test_data/SPICE` directory using `filecache`; these are downloaded into the same cache, distinct from the one used by spicedb. Retrieval is done in parallel for maximum download efficiency, and then the resultant local paths are supplied to `cspyce.furnsh` one at a time. Even tests that are not currently operational have been modified for future use. All tests have been updated and pass.
  - Changed gold master tests to use `filecache`:
    - The `obspath` attribute in a "standard obs" is no longer an absolute path, but is relative to the prefix specified by the `OOPS_TEST_DATA_PATH` environment variable. This is required because it's never permitted to access files outside of the roots defined by these variables.
    - Removed all uses of `os.path.join`, which doesn't play nicely with non-local-filesystem URIs. 
      - I also removed this from a few places in the hosts code, but not all of them. 
      - I also removed a lot of orphan `import os.path` statements that are no longer (or were never) needed to remind us not to use `os.path.join` in the future.
    - Removed all `makedirs` calls, since directories are now automatically created as necessary by `filecache`.
[...]

PR #146, opened 2024-11-26
Updates for new FCPath

### LINKED ISSUE #143: Update hosts to use filecache
Right now only the test system and spicedb have been updated to use filecache. The hosts modules should be changed as well. Functions like `from_file` should be able to take a URI and download and cache the files locally as necessary. This probably includes downloading both the image file and the detached label.

### LINKED ISSUE #144: Update tests and spicedb to use filecache Path-like functionality
Once implemented in filecache, change the tests and spicedb to use new functionality that emulates Path-like objects (allowing (a/"b"/"c").retrieve(), for example).

### LINKED ISSUE #145: Modify oops and hosts to use filecache FCPaths when they are implemented
When the `FCPath` class is implemented in `rms-filecache`, modify oops to use it.

rfrenchseti:
- Fixes #143 
- Fixes #144 
- Fixes #145 

Major changes:
- Change gold master tests to use new `FCPath` functionality from the `filecache` package. This returns much of the code to the way it looked before PR #138 and makes the code easier to read.
- Change all hosts to accept an `FCPath` as the argument to `from_file` and `from_index`. If the provided path is remote, the file is retrieved before use.
- Change `spicedb` to use `FCPath`.

Issues:
- Due to the lack of sufficient tests, most of the changes to the hosts modules have not been tested.


jnspitale (approved):
All units tests pass, looks nice.  Only comment I have is really about filecache: I see file retrieval is done explicitly.  It seems like this could be hidden inside the class, where the local path is stored inside the class instance and retrieval would be performed if the local path is not yet defined.

markshowalter (approved):
I will need to gain some experience with the new module, but I see no problem with your changes. Go ahead and merge.

PR #154, opened 2025-02-19
Hacks to fix hosts for navigation work

rfrenchseti:
These changes are required to make the hosts module not crash and/or just work for my navigation project. In most cases bad code is simply commented out and will need to be replaced with the correct code.

- Cassini ISS:
  - Turn off strict VICAR label processing.
  - Turn off NAV/WAC offset since this is left over from the old navigation project and is probably wrong.
- NH LORRI:
  - Comment out all code related to calibration. This is for two reasons: 1) the oops calibration classes have changed, and 2) the spectral headers like `RSOLAR` don't actually seem to be in the FITS header.
- Voyager ISS:
  - Comment out all code related to calibration. This is because the oops calibration classes have changed.
  - Work around a bug in the `vicar` library where the "TASK" label value is incorrect.

### DIFF
--- a/oops/hosts/cassini/iss.py
+++ b/oops/hosts/cassini/iss.py
@@ -39,7 +39,7 @@ def from_file(filespec, fast_distortion=True,
     # Load the VICAR file
     filespec = FCPath(filespec)
     local_path = filespec.retrieve()
-    vic = vicar.VicarImage.from_file(local_path)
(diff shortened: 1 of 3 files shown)


jnspitale (commented):
Changes look good.  Unit tests for hst, keck, jiram, and junocam are failing, though.

jnspitale (approved):
Rob says the unit tests I mentioned are not meant to work right now, so all good otherwise.

PR #161, opened 2025-02-28
New backplanes

jnspitale:
Mostly updates for the new backplanes.  There is also a fix to the Galileo SSI module to properly recognize  a null CUT_OUT_WINDOW.

I don't know why there is a mod to nircam.  So maybe that shouldn't be there.


oops/hosts/galileo/ssi/__init__.py
```diff
@@ -214,14 +214,18 @@ def __init__(self, meta_dict):
...
-            self.window_shape = self.window[2:]
-            self.window_uv_origin = np.flip(self.window_origin)
-            self.window_uv_shape = np.flip(self.window_shape)
-        else:
-            self.window = None
+            window = np.array(meta_dict['CUT_OUT_WINDOW'])
+            
+            # check for [-1,-1,-1,-1].  This is the value written in the 
+            # supplemental index when there is no CUT_OUT_WINDOW in the label.
+            if np.count_nonzero(window+1) != 0:
```
markshowalter, line 223: I find this line a bit obscure. How about `if np.all(window == -1)` instead? That's much more explicit about what you are testing.

oops/hosts/jwst/nircam/__init__.py
```diff
@@ -196,6 +196,12 @@ def from_hdulist(hdulist, **options):
             subfields['i_over_f'] = cal
 
         subfields['texp'] = subfields['headers'][0]['EFFEXPTM']
+
```
markshowalter, line 199: This looks like some old test code that I left behind in a previous checkin. Please delete.

oops/backplane/pixel.py
```diff
@@ -0,0 +1,70 @@
...
+    # compute apparent enclosing radius
+    if radius_in_km==0:
+        (body, mod) = Backplane.get_body_and_modifier(gridless_key[1])
+        radius_in_km = body.radius
+    radius = radius_in_km/distance / self.obs.fov.uv_scale.values[0]
+
+    return self.register_backplane(key, radius)
+
+#===============================================================================
+def _center_coordinates(self, gridless_key):
```
markshowalter, line 37: I'm not sure I see the need for this extra function. I suggest you just place these two lines inside the definition of `center_coordinate`.

markshowalter (changes requested):
This is good code but I have a bunch of comments, as usual.

oops/backplane/ring.py
```diff
@@ -653,6 +654,81 @@ def ring_radial_resolution(self, event_key):
 
     return self.register_backplane(key, resolution)
 
+#===============================================================================
+def _longitudinal_resolution(self, event_key, format="deg"):
```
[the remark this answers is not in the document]
pds-admin: `format` is not a reserved word in Python. It's a method of the `str` class. There's no conflict.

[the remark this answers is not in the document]
rfrenchseti: Sorry that comment was from me

oops/backplane/ring.py
```diff
@@ -653,6 +654,81 @@ def ring_radial_resolution(self, event_key):
...
+        return self._remasked_backplane(key, backplane_key)
+
+    if key in self.backplanes:
+        return self.get_backplane(key)
+
+    resolution = self._longitudinal_resolution(event_key)
+    return self.register_backplane(key, resolution)
+
+#===============================================================================
+def ring_longitudinal_resolution_km(self, event_key):
```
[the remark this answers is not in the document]
jnspitale: switched to ring_angular_resolution, added format keyword because it's more clear in the column descriptions.

oops/hosts/galileo/ssi/__init__.py
jnspitale: changed to `if np.all(window != -1)`

jnspitale:
Just pushed new commit with the requested changes.

oops/backplane/pixel.py
```diff
@@ -0,0 +1,65 @@
...
+    if radius==0:
+        (body, mod) = Backplane.get_body_and_modifier(gridless_key[1])
+        radius = body.radius
+    radius = radius/distance / self.obs.fov.uv_scale.values[0] ### only good for snapshots
+
+    return 2*self.register_backplane(key, radius)
+
+#===============================================================================
+def center_coordinate(self, event_key, axis="x"):
+    """Gridless v coordinate of the center of the disk.
```
markshowalter, line 39: I think you mean "v" to be "x or y" in the docstring.

oops/backplane/ring.py
```diff
@@ -654,21 +654,22 @@ def ring_radial_resolution(self, event_key):
...
-def ring_angular_resolution(self, event_key):
+def ring_angular_resolution(self, event_key, format="deg"):
     """Projected angular resolution in radians/pixel at the ring intercept.
 
     Input:
         event_key       key defining the ring surface event. Alternatively, a
                         ring_radius or radial_mode backplane key, in which case
                         this backplane inherits the mask of the given backplane
                         array.
+        format          longitude representation; "deg" or "km"
```
markshowalter, line 665: Please don't call this "format", as that is a reserved word. Also, "deg" is not right because the values are radians. If you plan to stick with a string value options for the input, rather than a boolean, be sure to add a check at the beginning and raise ValueError if the input is not one of the valid options.

oops/hosts/galileo/ssi/__init__.py
```diff
@@ -214,14 +214,18 @@ def __init__(self, meta_dict):
...
-            self.window_shape = self.window[2:]
-            self.window_uv_origin = np.flip(self.window_origin)
-            self.window_uv_shape = np.flip(self.window_shape)
-        else:
-            self.window = None
+            window = np.array(meta_dict['CUT_OUT_WINDOW'])
+            
+            # check for [-1,-1,-1,-1].  This is the value written in the 
+            # supplemental index when there is no CUT_OUT_WINDOW in the label.
+            if np.all(window != -1):
```
markshowalter, line 223: Is there ever an occasion where -1 appears in CUT_OUT_WINDOW without all four values being -1? Because this will not execute if there's a single occurrence of -1. Just making sure that's what you mean. If the only possible case is (-1,-1,-1,-1), then the most painfully explicit thing to write would be
```
if window != (-1,-1,-1,-1):
```
Also, what do you want to happen if the CUT_OUT_WINDOW really is (-1,-1,-1,-1)? Right now, the attributes are not filled in. Is (-1,-1,-1,-1) really equivalent to (1, 800, 1, 800)?

markshowalter (changes requested):
Scattered remaining comments.

oops/backplane/pixel.py
jnspitale: Actually  v is just spurious.  I don't think the "x or y" is needed.

oops/backplane/ring.py
jnspitale: changed to "units"

oops/hosts/galileo/ssi/__init__.py
jnspitale: I'm pretty sure this never happens the GLL SSI data images (but I'm checking).  Either all the elements are >=0 or they are all -1.  In the case of (-1,-1,-1,-1), self.window is left as None, and the image does not get cropped.

oops/backplane/pixel.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # compute apparent distance
+    event = self.get_surface_event(gridless_key, arrivals=True)
+    distance = event._dep_lt_*C
+
+    # compute apparent enclosing radius
+    if radius==0:
+        (body, mod) = Backplane.get_body_and_modifier(gridless_key[1])
+        radius = body.radius
+    radius = radius/distance / self.obs.fov.uv_scale.values[0] ### only good for snapshots
```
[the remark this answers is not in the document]
jnspitale: Updated to use "u" and "v", which seem more appropriate.

oops/hosts/galileo/ssi/__init__.py
jnspitale: I have confirmed that there are no cases in GLL SSI where CUT_OUT_WINDOW contains any -1 values unless all values are -1.  Changed to  

> if window.tolist() != [-1,-1,-1,-1]

to exactly correspond to the comment.

oops/backplane/ring.py
```diff
@@ -654,21 +654,24 @@ def ring_radial_resolution(self, event_key):
...
 
     Input:
         event_key       key defining the ring surface event. Alternatively, a
                         ring_radius or radial_mode backplane key, in which case
                         this backplane inherits the mask of the given backplane
                         array.
+        units           longitude representation; "rad" or "km"
     """
 
+    assert units.lower() in ["rad", "km"]
```
markshowalter, line 668: "assert" doesn't accomplish what we want. Please do the check and raise ValueError. ValueError is defined to be the error raised by any Python function in which the user passes an invalid or incorrect value. From https://docs.python.org/3/library/exceptions.html:

exception ValueError
Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

oops/backplane/pixel.py
```diff
@@ -0,0 +1,83 @@
...
+    distance = event._dep_lt_*C
+
+    # compute apparent enclosing radius
+    (body, mod) = Backplane.get_body_and_modifier(gridless_key[1])
+    if radius==0:
+        radius = body.radius
+
+    theta = radius/distance
+    radii_in_pixels = np.divide(theta, self.obs.fov.uv_scale.values)
+    if index:
```
markshowalter, line 48: This fails for index == 0, so it's not what you want.

markshowalter:
I know this isn't ready for review, but I have two quick comments.

- It's inconsistent to use `axis` = "u" or "v" for `body_diameter_in_pixels()` but `axis` = "x" or "y" in `center_coordinates()`. I think the latter is preferred, because this is a user-facing function and users are more likely to know what "x" and "y" mean. The u/v notation is mostly just internal to the source code.
- `assert` statements are for internal self-checks in the code, not for user-facing tests. ValueError is defined in the Python standard as the exception to raise when a user passes an invalid input to a function. So please change your assert statements to tests that raise ValueError with an explicit message.

jnspitale:
The asserts should already be gone.

jnspitale:
I would advocate changing to uv in center_coordinates.  OPUS users only see the column names, which are written as X and Y.

markshowalter (changes requested):
See comments.

markshowalter (approved):

Issue #153, opened 2025-02-12
NHLORRI is not working


rfrenchseti:
NHLORRI is giving negative resolutions, and backplane generation is also failing.

```
>>> import oops.hosts.newhorizons.lorri
>>> x=oops.hosts.newhorizons.lorri.from_file('https://opus.pds-rings.seti.org/holdings/volumes/NHxxLO_xxxx/NHPELO_2001/data/20150625_029751/lor_0297516223_0x633_sci.fit')
>>> inv=x.inventory(['PLUTO'], return_type='full')
>>> inv['PLUTO']['resolution']
array([ 455.27883432, -455.27882822])
```

jnspitale:
The reason it is returning a negative v scale is that lorri.py explicitly enters it as negative on line 478:

        full_fov_flat = oops.fov.FlatFOV((4.96e-6,-4.96e-6), (1024,1024))

Easily removed if we're sure there not some other reason for this.  I can't think of one.

jnspitale:
Actually, that does not fix the problem, so I'm leaving this open.

rfrenchseti:
(Reposting as correct user)

It would be really helpful for this to be addressed soon. I'd like to be able to say we have NH working in my annual status report.

jnspitale:
Proposed fix was to reverse the signs related to the Owen distortion model and flip the C matrix y axis.   However Mark pointed out that v_scale is allowed to be negative to account for the image sample direction. Therefore, the original host module is correct and the issue is that bounding boxes in the inventory are being generated with negative sizes.

jnspitale:
From Rob:
I think it's a lot worse than just the inventory. When I create a backplane using the old LORRI code (working around the inventory problem) I don't get anything useful.

jnspitale:
I'm getting reasonable backplanes for limb_clock_angle, right_ascension, and phase_angle for lor_0297516223_0x633_sci.fit.  Can you give me an example that fails?

rfrenchseti:
My mistake. The backplanes are fine. The problem is that v_min and v_max are swapped in the inventory.

PR #165, opened 2025-03-25
Nh lorri resolution fix

jnspitale:
Addresses #153.  The LORRI module was returning negative resolutions.  The cause was a spurious minus sign in the definition.


rfrenchseti:
I think this PR is no longer needed, right?

jnspitale:
That's right

Issue #186, opened 2025-11-17
Use pdsparser.PdsLabel.from_file() instead of pds3.get_label() in host modules


jnspitale:
Change:
`label = pds3.get_label(filespec)`
to:
`label = pdsparser.PdsLabel.from_file(filespec).as_dict()`

Issue #185, opened 2025-11-17
Move pds3.fast_dict calls to pdsparser


jnspitale:
uvis and vims currently call pds3.fast_dict.  The "fast" argument to pdsparser.pdsLabel() appears to implement the same functionality.

PR #191, opened 2026-02-11
Deprecate pds3.py

### LINKED ISSUE #184: Deprecate pds3.py
It looks like all of the functionality of pds3.py is now handled by pdsparser, so code that currently uses pds3.py should be changed to use pdsparser.

jnspitale:
Fixes #184, #185, and #186

## Summary by CodeRabbit

* **Refactor**
  * Updated label parsing across Cassini and Galileo instrument modules to use a new parsing approach.

* **Bug Fixes**
  * Fixed an iteration bug in Cassini UVIS that could skip processing of windowed data.

* **Chores**
  * Added test framework import to support unit testing.


rfrenchseti (approved):

markshowalter (approved):
One optional comment is that, in general, you should probably use `method='fast'` by default, and only abandon it if you discover empirically that it doesn't always work for some data set. But that's up to Joe.

PR #195, opened 2026-05-06
Update goldmaster to use filecache.replace() and remove explicit retrieve()

jnspitale:
Addresses #145 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved log and summary rotation to use atomic, timestamped backups based on stored file metadata, reducing failures and ensuring reliable archival of prior files.

* **Chores**
  * No changes to public interfaces or exported signatures.


jnspitale:
@coderabbitai review

oops/gold_master/__init__.py
```diff
@@ -1147,31 +1147,20 @@ def run_tests(self):
...
 
-            localpath = None
-            try:
-                localpath = log_path.retrieve()
-            except FileNotFoundError:
-                pass
-
-            if localpath:
+            localpath = log_path.get_local_path()
+            if os.path.exists(localpath):
```
rfrenchseti: I don't see how this could work. All this does is check to see if the file exists locally in your cache. It could just as well exist in a cloud bucket having been written by someone else (or you in a previous session and your cache was cleared since then). If the file exists in the cloud, it still needs to be renamed. Maybe you want to use `log_path.exists()` instead?

rfrenchseti (changes requested):

oops/gold_master/__init__.py
jnspitale: I think this preserves the original functionality, but not sure it's really the desired functionality.

rfrenchseti (approved):

PR #201, opened 2026-08-10
Add custom C-matrix support via generic Observation.get_cmatrix/set_cmatrix

jnspitale:
## Summary
- `hosts/cassini/iss.py` `from_file()` accepts a caller-supplied `cmatrix=` — the spice-convention C-matrix (J2000 → `CASSINI_ISS_<camera>`, exactly as returned by `cspyce.pxform()` or recorded in a CK) — plus an optional `frame_id=`. A custom-cmatrix load never loads a CK, never builds the SPICE camera frames (now built lazily by `ISS.define_camera_frames()`), and never reports CKs in `spice_kernels`.
- Generic `Observation.get_cmatrix(uv=None, time=None)` and `Observation.set_cmatrix(cmatrix, frame_id=None)` on the base class. Both are driven by a single host class attribute, `CMATRIX_ROTATION` (the spice-frame → oops-frame convention rotation, reached through the observation's `host` subfield); the inverse direction is derived by transposition. No host-specific set/get methods, and nothing frame-shaped is stored on the observation — `get_cmatrix` derives the C-matrix from `self.frame` at call time.
- `set_cmatrix` validates its input: the matrix must be a proper floating-point rotation. Non-finite values, boolean/integer data, reflections, scalings, and non-orthogonal matrices raise `ValueError` at the boundary.
- `set_cmatrix(frame_id=)` only overrides frame registrations it made itself; an ID already held by any other frame (e.g. `CASSINI_ISS_NAC`, `J2000`) raises instead of silently replacing its primary definition. `frame_id=None` (default) gives each observation its own unregistered frame; re-using a set_cmatrix-issued ID re-points every observation sharing that registered frame, which is the point of giving it an ID.
- `oops/frame/cmatrix.py` `Cmatrix` gains an `override=` option so an owned registration can be cleanly replaced.

## Breaking changes
- `CASSINI_ISS_NAC_FLIPPED` / `CASSINI_ISS_WAC_FLIPPED` are renamed to `*_SPICE`.
- `ISS.initialize()` alone no longer registers `CASSINI_ISS_NAC`/`CASSINI_ISS_WAC`; they appear after the first plain (SPICE-pointed) `from_file()`, which calls `define_camera_frames()`.
- The `offset_wac` initialize option is removed: it made the WAC pointing a composite (Navigation offset on top of the SPICE chain), so `get_cmatrix` could not agree with `pxform` and the meaning of `from_file(cmatrix=)` depended on an initialize-time flag.

## Review
Addresses the independent verification review (Custom C-Matrix Verification, R. French, 2026-08-14): F1 rotation validation, F2 frame-registry protection, F3 `offset_wac` removal, F5/F6 docstring clarifications (stale derived geometry; no CK loaded so `pxform` against the SPICE camera frame fails for that epoch). F4 (enabling the other hosts: Galileo SSI, Voyager ISS, NH LORRI) is deliberately deferred to #204 — this branch is the template.

## Test plan
- [x] `tests/hosts/cassini/iss/test_iss.py` — 18 tests: round-trip vs `pxform`, per-observation frame isolation, no leakage into plain loads, frame_id reuse/collision, C-matrix validation, CK loading and `spice_kernels` reporting
- [x] `tests/frame/test_cmatrix.py`
- [x] `tests/observation` suite

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015Wg9zMfz6FNEotebnS1hmm

### DIFF
--- a/oops/frame/cmatrix.py
+++ b/oops/frame/cmatrix.py
@@ -21,7 +21,7 @@ class Cmatrix(Frame):
     # expendable. Frame IDs are not preserved during pickling.
 
     #===========================================================================
-    def __init__(self, cmatrix, reference=None, frame_id=None):
+    def __init__(self, cmatrix, reference=None, frame_id=None, override=False):
         """Constructor for a Cmatrix frame.
 
         Input:
@@ -30,6 +30,8 @@ class Cmatrix(Frame):
                         None for J2000.
             frame_id    the ID under which the frame will be registered; None
                         to leave the frame unregistered
+            override    if True, replace the primary definition of any frame
+                        already registered under this frame_id.
         """
 
         self.cmatrix = Matrix3.as_matrix3(cmatrix)
@@ -42,7 +44,7 @@ class Cmatrix(Frame):
         self.keys      = set()
 
         # Update wayframe and frame_id; register if not temporary
-        self.register()
+        self.register(override=override)
 
         # It needs a wayframe before we can construct the transform
         self.transform = Transform(cmatrix, Vector3.ZERO,
--- a/oops/hosts/cassini/__init__.py
+++ b/oops/hosts/cassini/__init__.py
@@ -350,9 +350,21 @@ class Cassini(object):
 
     #===========================================================================
     @staticmethod
-    def used_kernels(time, inst, return_all_planets=False):
+    def used_kernels(time, inst, return_all_planets=False, ck=True):
         """The list of kernels associated with a Cassini observation at a
         selected range of times.
+
+        Input:
+            time                a (start, stop) tuple of times in seconds TDB.
+            inst                the instrument name, e.g., 'iss'.
+            return_all_planets  Include kernels for all planets not just
+                                Jupiter or Saturn.
+            ck                  True (default) to include CK (pointing)
+                                kernels; False to exclude them, e.g. for an
+                                observation whose pointing came from a custom
+                                cmatrix rather than SPICE, where any
+                                furnished CK is unrelated to how the
+                                observation was actually pointed.
         """
         if return_all_planets:
             bodies = [1, 199, 2, 299, 3, 399, 4, 499, 5, 599, 6, 699,
@@ -367,7 +379,11 @@ class Cassini(object):
             else:
                 bodies = [5, 599] + Body.JUPITER_MOONS_LOADED
 
-        return spicedb.used_basenames(time=time, inst=inst, sc=-82,
-                                      bodies=bodies)
+        types = None
+        if not ck:
+            types = [t for t in spicedb.KERNEL_TYPE_SORT_ORDER if t != 'CK']
+
+        return spicedb.used_basenames(types=types, time=time, inst=inst,
+                                      sc=-82, bodies=bodies)
 
 ################################################################################
--- a/oops/hosts/cassini/iss.py
+++ b/oops/hosts/cassini/iss.py
@@ -31,7 +31,7 @@ CMATRIX_ROTATION = oops.Matrix3([[-1,0,0],[0,-1,0],[0,0,1]])
 
 def from_file(filespec, fast_distortion=True,
               return_all_planets=False, cmatrix=None, frame_id=None,
-              map_other_camera=False, **parameters):
+              **parameters):
     """A general, static method to return a Snapshot object based on a given
     Cassini ISS image file.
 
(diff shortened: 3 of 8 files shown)


jnspitale:
@coderabbitai review

astrocfi:
See report: https://claude.ai/code/artifact/ae1c9373-6d52-4efa-878d-2943637ef4f4

rfrenchseti:
One forward-compatibility note on `Observation.set_cmatrix()`, since #203 is in flight against the same base.

The `frame_id=None` branch is documented as attaching "a fresh unregistered frame owned by the observation, so loading other images never disturbs its pointing". That ownership property does hold as this PR stands — I checked that constructing many unregistered `Cmatrix` frames leaves `Frame.FRAME_CACHE` and `Frame.WAYFRAME_REGISTRY` flat at 2 entries, and that two frames built from an equal-valued matrix are distinct objects.

But it isn't a property that `Cmatrix.__init__` or `Frame.register()` actually promises. It is emergent from how the current registry treats a `frame_id` of `None`, and #203 changes exactly that: under its registry rewrite `Frame._register()` inserts every frame into a per-subclass `_WAYFRAMES` dict regardless of `frame_id`, and equal-valued unregistered frames dedup to a single shared wayframe object. On that branch `Cmatrix(m)._wayframe is Cmatrix(m)._wayframe` is `True`, and each construction is retained globally for the life of the process.

If #203 lands after this, the docstring's guarantee becomes false without anything in either PR failing — the frame is no longer owned by the observation, and it is no longer isolated from global state. Two observations handed equal-valued C-matrices would share a frame object.

Either of these would be enough to keep that from happening quietly:

- Document the weaker, implementation-independent property — that the frame is simply not registered under a frame ID — rather than ownership or isolation from other loads.
- Add a test that pins the property you actually want: that two observations given equal-valued C-matrices don't end up sharing a frame object, and that `set_cmatrix(frame_id=None)` leaves the frame registry unchanged. That turns a silent semantic regression into a failing test whenever the registry internals move.

Worth coordinating with #203 either way, since the two PRs are independent branches off `main` and whichever merges second inherits the interaction.

rfrenchseti (changes requested):
See earlier comments

jnspitale:
Good catch — agreed that the isolation property was emergent rather than promised anywhere. Addressed by pinning it as a test rather than weakening the docstring, since per-observation ownership is the feature this argument exists for, not an accident we're willing to lose:

`test_default_frames_are_distinct_and_stay_out_of_registry` now asserts exactly the two properties you suggested — observations given equal-valued C-matrices get distinct frame objects, and a `set_cmatrix(frame_id=None)` load leaves `Frame.WAYFRAME_REGISTRY` and `Frame.FRAME_CACHE` at their prior sizes (measured after a warm load+`get_cmatrix` cycle, so lazy one-time registrations don't confound it). Under the #203 registry semantics you describe, both assertions fail, so whichever branch merges second now hits a failing test at rebase instead of a silent regression.

The retention/dedup question itself is #203's design call — raised there with a pointer back to this contract.

jnspitale:
This is ready for re-review.  Main concern is interaction with 203.

markshowalter (changes requested):
We're going to need to discuss this. My first issue is that the `override` option in `Cmatrix` needs to be implemented via the new `Fittable/Mutable` API. You can't just replace the definition of an object without notifying any object on which it depends. This is what the new API handles automatically. I did not define class `Cmatrix` to be `Fittable`, and recommend that we don't. Instead, a user should be able to define a `Navigation` frame that operates on the `Cmatrix`, and then adjust the `Navigation`. A lot of this needs to be built into the planned rewrite of the `host` modules. Until then, we can discuss what is specifically needed as an interim fix. 

_Maybe_ we can just define every Cassini frame as a `Navigation` atop a `Cmatrix`, because a `Navigation` object has very little overhead, and then you could update the `Navigation` at will using `obs.frame.set_params()`. That would guarantee that the changes propagate forward to anything derived from `obs`, such as backplanes. However, the quantities you use to define the `Navigation` are 2 or 3 rotation angles, not  a brand new `Cmatrix`.

My only issue with `set_cmatrix` is the name, because we use "`cmatrix`" widely in the source code to refer to the OOPS version, whereas this function wants to return the SPICE C kernel definition, whatever it is. The solution would be to rename it `get_spice_cmatrix`, so there is no ambiguity. Eventually, `get_spice_cmatrix` will need to be implemented for all hosts, but that can come with the `hosts` cleanup.

jnspitale:
I also do not like using cmatrix when we're talking about an orientation in the OOPS convention.  I use "orientation matrix" to denote this in OMINAS and reserve "cmatrix" for the SPICE orientation matrix.  This would be less confusing.

rfrenchseti:
I'm assuming this should be closed now since it's been superseded by #203.

rfrenchseti:
Superseded by #203

